PR #144, opened 2026-02-28
feat(cycle-037): Agent-Native Output Protocol

zkSoju:
## Summary

Agent-Native Output Protocol for the Loa Constructs Network. Introduces TOON (Token-Oriented Object Notation) encoding, per-invocation CTA protocol, hash-based divergence detection, and lazy-loading contract documentation.

### Changes

- **TOON encoder** (`toon-lib.sh`): Header+CSV tabular format achieving ~39.6% token reduction for agent consumers
- **4-arg format router** (`constructs-lib.sh`): `format_tabular_output(label, tabular_json, original_payload, fallback_fn)` — separates data shapes to prevent fallback mismatch
- **CTA protocol** (`skill-cta.md`): Per-invocation `Next:` block with up to 3 context-sensitive command suggestions
- **Hash divergence detection** (`constructs-loader.sh`): Reports `[DIVERGED]` when version matches but content hash differs
- **Lazy-loading contract** (`runtime-contract.md`): Formal specification — skills auto-load on invocation, only index.yaml at session start
- **`workflow_next` field** (`types.ts`, `validation.ts`): Cross-construct navigation hints in PackManifest
- **Shared status data collection** (`constructs-install.sh`): Single collection pass feeds both TOON and markdown rendering paths
- All new behaviors default to OFF — config-gated, additive-only

### Bridge Review Summary

| Metric | Value |
|--------|-------|
| Iterations | 3 (2 reviewed + 1 flatline) |
| Findings discovered | 5 (2 MEDIUM, 2 LOW, 1 cosmetic) |
| Findings addressed | 3 |
| Findings deferred | 2 (by design) |
| PRAISE findings | 4 |

**Iteration 1**: Identified duplicated HTTP fetches in TOON status path, missing divergence detection in non-markdown output, and toon-lib re-sourcing overhead.

**Iteration 2**: All 3 actionable findings resolved — shared data collection (Monarch pattern), registry hash fetch added, source guard implemented. 0 new actionable findings.

**Iteration 3**: Flatline confirmed — no code changes since iteration 2.

### Files Changed

15 files (+1,465 / -662)

### Test Plan

- [x] Default config produces byte-identical output (no behavior change)
- [x] `output.format: toon` produces TOON-encoded tabular output
- [x] `cta.enabled: true` appends Next: block after status/install
- [x] Hash divergence detection reports [DIVERGED] correctly
- [x] TypeScript compilation clean (shared package)
- [x] TOON fallback on encoding failure produces markdown output

---

🤖 Generated with [Claude Code](https://claude.com/claude-code) | Bridgebuilder reviewed

### DIFF
--- a/.claude/protocols/skill-cta.md
+++ b/.claude/protocols/skill-cta.md
@@ -1,0 +1,44 @@
+# Skill CTA Protocol
+
+> **Source**: cycle-037 SDD §2.4, PRD FR-2.1
+> **Status**: Active (cycle-037 MVP)
+
+## Format
+
+After primary command output, CTA-enabled commands append:
+
+```
+Next:
+<command-1> — <description>
+<command-2> — <description>
+```
+
+## Rules
+
+1. Maximum 3 CTAs per invocation
+2. CTAs appear after main output, before any grimoire writes
+3. Use truenames (e.g., `/implement`) not golden path aliases (e.g., `/build`)
+4. Command-context-based: each CTA-enabled command has a static mapping of relevant next steps (cycle-037 MVP). Dynamic workflow-state-derived CTAs using `golden_path.commands` + `truename_map` are future scope.
+5. Gated by `cta.enabled: true` in `.loa.config.yaml`
+
+## Enabled Commands (cycle-037 MVP)
+
+| Command | Context | CTAs |
+|---------|---------|------|
+| `constructs browse` | browse | install, status, loa |
+| `constructs status` | status | browse, install, loa |
+| `constructs install` | install | quick_start (if available), status, loa |
+
+## Configuration
+
+```yaml
+# .loa.config.yaml
+cta:
+  enabled: false  # default: false
+```
+
+## Implementation
+
+- `emit_cta()` in `constructs-lib.sh`
+- `is_cta_enabled()` in `constructs-lib.sh`
+- Protocol file: `.claude/protocols/skill-cta.md` (this file)
--- a/.claude/scripts/constructs-browse.sh
+++ b/.claude/scripts/constructs-browse.sh
@@ -292,10 +292,26 @@ cmd_list() {
     fi
 
     if [[ "$json_output" == true ]]; then
+        # Explicit --json flag always wins (backwards compat)
         format_packs_json "$packs_json"
     else
-        format_packs_human "$packs_json"
+        # Build flat tabular JSON for TOON/json modes
+        local toon_json
+        toon_json=$(echo "$packs_json" | jq '[
+            .data[]? | {
+                slug: .slug,
+                name: .name,
+                skills: (.skills_count // (.manifest.skills | length?) // 0),
+                version: (.latest_version.version // .version // "1.0.0"),
+                tier: (.tier_required // .tier // "free")
+            }
+        ]' 2>/dev/null)
+        # Route: toon→TOON encoder, json→raw JSON, md→format_packs_human
+        format_tabular_output "packs" "$toon_json" "$packs_json" "format_packs_human"
     fi
+
+    # Append CTAs if enabled
+    emit_cta "browse"
 }
 
 cmd_info() {
--- a/.claude/scripts/constructs-install.sh
+++ b/.claude/scripts/constructs-install.sh
@@ -809,6 +809,9 @@ PYEOF
         done
     fi
 
+    # Append CTAs if enabled (cycle-037)
+    emit_cta "install" "$pack_slug"
+
     return $EXIT_SUCCESS
 }
 
@@ -2015,7 +2018,7 @@ do_status_pack() {
     registry_url=$(get_registry_url)
 
     if [[ -n "$pack_slug" ]]; then
-        # Single pack status
+        # Single pack status — always use detailed format
         show_pack_status "$pack_slug" "$meta_path" "$registry_url"
     else
         # All installed packs
@@ -2026,17 +2029,85 @@ do_status_pack() {
             return $EXIT_SUCCESS
         fi
 
-        echo "Construct Status"
-        echo "════════════════════════════════════════"
-        echo ""
+        local fmt
+        fmt=$(get_output_format)
 
-        while IFS= read -r slug; do
-            show_pack_status "$slug" "$meta_path" "$registry_url"
+        if [[ "$fmt" == "toon" || "$fmt" == "json" ]]; then
+            # Collect all pack data into JSON array for TOON/json routing
+            local status_json="[]"
+            while IFS= read -r slug; do
+                local lv rv sl
+                lv=$(jq -r ".installed_packs[\"$slug\"].version // \"unknown\"" "$meta_path" 2>/dev/null)
+                local lh
+                lh=$(jq -r ".installed_packs[\"$slug\"].content_hash // \"\"" "$meta_path" 2>/dev/null)
+
+                # Fetch registry version (non-blocking)
+                rv="unknown"
+                local tmp_file
+                tmp_file=$(mktemp)
+                chmod 600 "$tmp_file"
+                local http_code
+                http_code=$(curl -s -w "%{http_code}" \
+                    --proto =https --tlsv1.2 --max-time 10 \
+                    "${registry_url}/constructs/${slug}" \
+                    -o "$tmp_file" 2>/dev/null) || http_code="000"
+                if [[ "$http_code" == "200" ]]; then
+                    rv=$(jq -r '.data.version // .data.latest_version.version // "unknown"' "$tmp_file" 2>/dev/null)
+                fi
+                rm -f "$tmp_file"
+
+                # Determine status label
+                sl="UNKNOWN"
+                if [[ -z "$lh" ]]; then
+                    if [[ "$lv" == "$rv" && "$rv" != "unknown" ]]; then sl="UNKNOWN"; else sl="BEHIND"; fi
+                else
+                    sl="SYNCED"
+                    if [[ "$lv" != "$rv" && "$rv" != "unknown" ]]; then sl="BEHIND"; fi
+                fi
+
+                status_json=$(echo "$status_json" | jq --arg s "$slug" \
+                    --arg v "$lv" --arg rv "$rv" --arg st "$sl" \
+                    '. += [{"slug":$s,"local":$v,"registry":$rv,"status":$st}]')
+            done <<< "$slugs"
+
+            format_tabular_output "status" "$status_json" "$status_json" "_show_all_packs_md"
+        else
+            # Default markdown: use existing per-pack output
+            echo "Construct Status"
+            echo "════════════════════════════════════════"
             echo ""
-        done <<< "$slugs"
+
+            while IFS= read -r slug; do
+                show_pack_status "$slug" "$meta_path" "$registry_url"
+                echo ""
+            done <<< "$slugs"
+        fi
     fi
 
+    # Append CTAs if enabled
+    emit_cta "status"
+
     return $EXIT_SUCCESS
+}
+
+# Markdown fallback for format_tabular_output when showing all pack status
+# Args: $1 = JSON array of status objects (unused — re-renders from meta)
+_show_all_packs_md() {
+    local meta_path
+    meta_path=$(get_registry_meta_path)
+    local registry_url
+    registry_url=$(get_registry_url)
+
+    echo "Construct Status"
+    echo "════════════════════════════════════════"
+    echo ""
+
+    local slugs
+    slugs=$(jq -r '.installed_packs | keys[]' "$meta_path" 2>/dev/null)
+    while IFS= read -r slug; do
+        show_pack_status "$slug" "$meta_path" "$registry_url"
+        echo ""
+    done <<< "$slugs"
 }
 
 # Show status for a single pack
--- a/.claude/scripts/constructs-install.sh
+++ b/.claude/scripts/constructs-install.sh
@@ -2029,9 +2029,6 @@ do_status_pack() {
             return $EXIT_SUCCESS
         fi
 
-        local fmt
-        fmt=$(get_output_format)
-
         # Collect all pack data into JSON array (shared data collection)
         local status_json="[]"
         while IFS= read -r slug; do
--- a/.claude/scripts/constructs-install.sh
+++ b/.claude/scripts/constructs-install.sh
@@ -2032,55 +2032,74 @@ do_status_pack() {
         local fmt
         fmt=$(get_output_format)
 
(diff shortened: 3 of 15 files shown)


zkSoju:
# Bridgebuilder Review — cycle-037 Agent-Native Output Protocol

**PR**: #144 | **Branch**: feat/cycle-037-agent-native-output | **Iteration**: 1
**Files**: 15 changed (+1,435 / -662)

---

## Opening Context

There is a pattern that recurs in every system that survives long enough to matter: the system starts by speaking one language to all its consumers — humans, machines, scripts — and eventually discovers that polyglot fluency is not a luxury but a survival trait.

What we are looking at in this PR is the Constructs Network learning to speak a second language. The existing markdown output served human-readable CLI sessions well. But as agent-driven workflows become the primary consumer — 20+ tool invocations per session, each one paying the token tax — the cost of monolingual output compounds. TOON is not merely a new serialization format; it is an architectural acknowledgment that the primary reader of this output is no longer a human terminal, but a language model's context window.

The implementation is clean, well-bounded, and — critically — defaults to OFF. This is how you introduce a new encoding layer: behind a config gate, with a fallback path that preserves byte-identical behavior. Google's Protocol Buffers had the same discipline — wire format v1 and v2 coexisted for years before v1 was deprecated. You earn the right to break compatibility by first proving you don't need to.

---

## Architectural Observations

### The 4-Arg Router — An Elegant Solution to a Subtle Problem

The `format_tabular_output(label, tabular_json, original_payload, fallback_fn)` design is worth calling out specifically. The GPT 5.3-codex review caught the data-shape mismatch risk during SDD iteration — TOON encoding takes flat, uniform JSON, but the markdown fallback expects the original API envelope. By separating these into distinct arguments, the implementation makes it impossible for a TOON failure to pass incompatible data to the markdown formatter.

This is the same principle behind Stripe's API versioning: the wire format and the internal representation are distinct, and the translation layer never assumes they are interchangeable.

### CTA as Pull-Based Discovery

The CTA protocol is deceptively simple — static command suggestions appended after output. But the architectural insight is deeper: this transforms CLI output from "here is what happened" to "here is what happened, and here is what you can do next." Netflix's Zuul gateway evolved similarly — from a simple proxy to a context-aware router that could suggest alternative paths when the primary route was degraded.

The static mapping in the MVP is correct. Dynamic CTAs from `golden_path.commands` + `truename_map` would require per-invocation state inference, which is a different problem. Ship the static version, prove the pattern, then add intelligence.

### Lazy-Loading Contract — Documenting the Folklore

The Skill Loading Contract in `runtime-contract.md` codifies what was previously tribal knowledge. The token baseline table (session-start: ~2,000 tokens vs eager: ~15,000-100,000 tokens) is exactly the kind of quantified justification that runtime implementors need. The `defer_loading` section is correctly marked non-normative — it documents intent without creating obligation.

---

## Findings


```json
{
  "schema_version": 1,
  "bridge_id": "bridge-20260228-c037",
  "iteration": 1,
  "findings": [
    {
      "id": "medium-1",
      "title": "TOON status collection duplicates HTTP calls already made by show_pack_status()",
      "severity": "MEDIUM",
      "category": "performance",
      "file": ".claude/scripts/constructs-install.sh:2035-2071",
      "description": "When output format is toon/json, do_status_pack() makes O(n) HTTP calls to collect pack metadata into a JSON array. But _show_all_packs_md() (the fallback) also calls show_pack_status() which makes its own O(2n) HTTP calls. This means a TOON encoding failure results in O(3n) HTTP calls — the collection pass plus the fallback's full pass. While this is an edge case (TOON encoding rarely fails on uniform data), the duplication is architectural debt.",
      "suggestion": "Consider extracting the HTTP fetch into a shared data-collection step that both the TOON path and the markdown path consume. The status_json array could be the single source of truth, with _show_all_packs_md() rendering from it instead of re-fetching.",
      "faang_parallel": "Google's internal dashboards learned this lesson with Monarch — the monitoring system that separated data collection from data presentation, ensuring each metric was fetched exactly once regardless of how many views consumed it.",
      "teachable_moment": "When you add a new rendering path to existing code, check whether the data collection is shared or duplicated. The rendering layer should be a pure function over already-collected data."
    },
    {
      "id": "medium-2",
      "title": "Status TOON path does not fetch registry hash — divergence detection incomplete",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": ".claude/scripts/constructs-install.sh:2059-2066",
      "description": "In the TOON/json status collection loop, the status label determination only considers version comparison and local hash presence. It does NOT fetch the registry hash — so DIVERGED status can never appear in TOON/json output. The markdown path (via show_pack_status()) does fetch the registry hash and correctly shows DIVERGED. This means switching from md to toon output silently loses divergence detection.",
      "suggestion": "Add registry hash fetch to the TOON collection loop (same pattern as show_pack_status():2167-2181), and include a 'DIVERGED' check when local_hash and registry_hash both exist but differ.",
      "metaphor": "Like a dashboard that shows green for all services but only because it forgot to check the health endpoint for half of them.",
      "teachable_moment": "When adding a new rendering path for existing data, verify that the data collection is feature-complete relative to the original path. A diff between the two code paths reveals gaps."
    },
    {
      "id": "low-1",
      "title": "toon_encode_tabular sources toon-lib.sh on every invocation",
      "severity": "LOW",
      "category": "performance",
      "file": ".claude/scripts/constructs-lib.sh:1351-1353",
      "description": "format_tabular_output() runs `source \"$script_dir/lib/toon-lib.sh\"` every time it is called with toon format. For the current 2 integration points this is negligible, but as TOON adoption expands (beads-health, golden-path, loader), repeated sourcing adds unnecessary overhead.",
      "suggestion": "Add a guard variable (e.g., `[[ -n \"${_TOON_LIB_LOADED:-}\" ]] || { source ...; _TOON_LIB_LOADED=1; }`) to source once per process.",
      "teachable_moment": "Bash source is idempotent but not free — it re-parses the file each time. Guard variables are the bash equivalent of include guards in C."
    },
    {
      "id": "low-2",
      "title": "emit_cta install context CTA may exceed 3 maximum",
      "severity": "LOW",
      "category": "spec-compliance",
      "file": ".claude/scripts/constructs-lib.sh:1412-1427",
      "description": "The install context always emits 2 CTAs (status + loa), and conditionally emits a quick_start CTA. If quick_start exists, that's 3 CTAs total — exactly at the limit. But if the quick_start command itself contains multiple lines (unlikely but possible if construct.yaml is malformed), the output could exceed 3 lines.",
      "suggestion": "No immediate change needed — construct.yaml quick_start.command is validated as a single string. But consider adding a line count check or `head -1` on quick_cmd for defense in depth.",
      "teachable_moment": "When a protocol says 'maximum N', consider both the happy path (where N is respected by construction) and the adversarial path (where upstream data could violate assumptions)."
    },
    {
      "id": "praise-1",
      "severity": "PRAISE",
      "title": "Config-gated, additive-only architecture",
      "description": "Every new behavior defaults to OFF. The existing CLI produces byte-identical output with default config. This is textbook feature flag discipline — the kind that lets you ship fearlessly because the blast radius of a misconfigured flag is 'nothing changes' rather than 'everything breaks'.",
      "suggestion": "No changes needed — this is exemplary.",
      "praise": true,
      "teachable_moment": "The safest default is always 'do nothing new.' Feature flags should require explicit opt-in, not explicit opt-out."
    },
    {
      "id": "praise-2",
      "severity": "PRAISE",
      "title": "The 4-arg format_tabular_output() design prevents data shape mismatch",
      "description": "Separating tabular_json (TOON-shaped) from original_payload (fallback-shaped) as distinct function arguments makes it structurally impossible for a TOON failure to pass incompatible data to the markdown formatter. This was caught during GPT cross-model review and the fix is clean.",
      "suggestion": "No changes needed — this is the correct architecture.",
      "praise": true,
      "faang_parallel": "Stripe's API versioning maintains separate wire format and internal representation — the same principle of never assuming format interchangeability.",
      "teachable_moment": "When you have two consumers of the same data that expect different shapes, make the shapes explicit in the function signature rather than relying on runtime conversion."
    }
  ]
}
```


---

## Closing Reflections

This PR introduces a layered output architecture that respects both the present (markdown for human readers) and the future (TOON for agent context windows). The implementation discipline — config-gated, fallback-safe, additive-only — is exactly right for a format change that touches the output of core CLI commands.

The two MEDIUM findings are worth addressing: the duplicated HTTP fetches in the TOON status path create architectural debt that will compound as TOON adoption expands, and the missing divergence detection in TOON output is a correctness gap that could confuse users who switch output formats and get different status information.

Neither finding blocks the PR. Both are the kind of refinement that makes the difference between code that works and code that scales.

The TOON encoder itself is a clean, minimal implementation. Two functions, no dependencies beyond jq (already required), with proper fallback on every failure path. The CTA protocol is correctly scoped as a static MVP with a clear path to dynamic evolution. And the runtime contract addition transforms undocumented folklore into a formal specification that other runtimes can implement.

Good work. Ship it, then refine.

zkSoju:
# Bridgebuilder Review — cycle-037 Agent-Native Output Protocol

**PR**: #144 | **Branch**: feat/cycle-037-agent-native-output | **Iteration**: 2
**Files**: 2 changed (+80 / -50) — fix commits addressing iteration 1 findings

---

## Opening Context

The mark of a codebase that is learning to be excellent is not the absence of findings — it is the quality of the response to findings.

Iteration 1 identified two MEDIUM-severity architectural concerns: duplicated HTTP fetch paths that created O(3n) network calls on TOON fallback, and incomplete divergence detection that silently dropped the DIVERGED status when switching output formats. Both were the kind of bug that doesn't crash anything — they just make the system slightly less trustworthy, which is worse.

The fix commits address all three actionable findings (medium-1, medium-2, low-1) with a clean architectural refactoring that transforms the code from "two rendering paths that each collect their own data" to "one data collection path that feeds two renderers." This is the Monarch pattern in miniature — separate collection from presentation, and the problem of duplicated fetches becomes structurally impossible.

---

## Iteration 1 Finding Resolution

### medium-1: Duplicated HTTP calls — RESOLVED

The refactoring extracts data collection into a shared loop (lines 2032-2090) that runs unconditionally before format routing. Both the TOON/json path and the markdown path now consume the same `status_json` array. The fallback function `_show_all_packs_md()` delegates to `_render_status_md()` which renders from pre-collected data rather than re-fetching.

The HTTP call count is now O(2n) in all cases — n for version, n for hash — regardless of output format or fallback path. This is correct.

### medium-2: Missing registry hash fetch — RESOLVED

The shared collection loop now fetches the registry hash (lines 2055-2068) for every pack. The status determination logic (lines 2070-2085) includes the full three-way check: SYNCED when hashes match, BEHIND when versions differ, DIVERGED when versions match but hashes differ. This is feature-complete relative to the original `show_pack_status()` markdown path.

### low-1: toon-lib.sh sourced every invocation — RESOLVED

A `_TOON_LIB_LOADED` guard variable (constructs-lib.sh:1352-1356) prevents re-parsing on subsequent calls. This is the bash equivalent of a C include guard — idiomatic and correct.

### low-2: CTA max 3 — DEFERRED (by design)

The iteration 1 review explicitly noted "No immediate change needed." The construct.yaml schema validates `quick_start.command` as a single string, making the edge case impossible under normal operation. Deferral is the correct call — defense-in-depth against malformed YAML is a separate concern.

---

## Findings


```json
{
  "schema_version": 1,
  "bridge_id": "bridge-20260228-c037",
  "iteration": 2,
  "findings": [
    {
      "id": "low-3",
      "title": "_show_all_packs_md fallback re-derives meta_path and registry_url",
      "severity": "LOW",
      "category": "style",
      "file": ".claude/scripts/constructs-install.sh:2111-2118",
      "description": "_show_all_packs_md() is the fallback function passed to format_tabular_output(). When TOON encoding fails, this function is called with the original payload. It re-derives meta_path via get_registry_meta_path() and registry_url via get_registry_url() — both of which are already known in the calling scope of do_status_pack(). These are pure functions that return the same value, so correctness is not at risk, but the pattern is slightly wasteful.",
      "suggestion": "This is cosmetic. The fallback path is rare (TOON encoding on uniform JSON essentially never fails). The re-derivation adds microseconds. No change recommended unless the fallback function signature is refactored in a future iteration.",
      "teachable_moment": "When a fallback function needs context that the caller already has, you face a design choice: pass context via closure (cleaner but harder in bash), pass via globals (easier but less explicit), or re-derive (safest, slightly redundant). In bash, re-derivation is often the right call — bash closures don't exist, and globals create invisible coupling."
    },
    {
      "id": "praise-3",
      "severity": "PRAISE",
      "title": "The Monarch refactoring is textbook separation of collection and presentation",
      "description": "The shared data collection loop is the single source of truth for pack status. Both rendering paths (TOON/json and markdown) consume the same status_json array. The _render_status_md() helper is a pure function over pre-collected data. This makes it structurally impossible for the two output formats to show different status information — the very bug that medium-2 identified is now prevented by architecture, not by testing.",
      "suggestion": "No changes needed. This is the correct pattern.",
      "praise": true,
      "faang_parallel": "Google's Monarch monitoring system separates metric collection (one pipeline) from metric presentation (many dashboards). The collection pipeline is the single source of truth. Adding a new dashboard never adds a new collection path.",
      "teachable_moment": "The best fix for 'these two code paths show different data' is never 'make them show the same data' — it is 'make them consume the same data.' The former requires ongoing synchronization; the latter makes divergence structurally impossible."
    },
    {
      "id": "praise-4",
      "severity": "PRAISE",
      "title": "Clean commit hygiene — surgical fixes with clear provenance",
      "description": "The fix is split into two commits: one for the architectural refactoring (shared collection, hash fetch, source guard) and one for the cleanup of a leftover duplicate variable declaration introduced by the refactoring. Each commit message references the bridge finding it addresses. This is exactly how you want to see bridge iteration fixes land — traceable, atomic, and reviewable.",
      "suggestion": "No changes needed.",
      "praise": true,
      "teachable_moment": "When fixing review findings, resist the urge to bundle everything into one commit. Atomic commits that reference specific findings create a permanent audit trail. Future engineers can trace any line back to the review that motivated it."
    }
  ]
}
```


---

## Closing Reflections

This iteration confirms convergence. The three actionable findings from iteration 1 are all resolved — the two MEDIUM findings through architectural refactoring, the LOW finding through an idiomatic source guard. The single new finding (low-3) is cosmetic and explicitly deferred.

The code is now in a state where:
- Data collection happens exactly once, regardless of output format
- Divergence detection is feature-complete across all output paths
- The TOON library is sourced once per process
- Every fix commit traces back to the specific bridge finding that motivated it

The delta between iteration 1 and iteration 2 is minimal — one LOW cosmetic finding versus two MEDIUMs and a LOW. This is flatline territory. The code has converged.

Ship it.

PR #147, opened 2026-03-08
fix: cycle-038 migration, auth wiring, seed visibility

zkSoju:
## Summary

Fixes the triple-fault causing empty registry + auth 500s downstream after cycle-038 deployment:

- **Missing migration** (`0008_cycle_038_visibility.sql`): Creates enums, columns, indexes for visibility system. Backfills existing packs to `public`. Fully idempotent.
- **Auth wiring**: All 5 `generateTokens()` call sites now pass `githubOrgMember`. GitHub OAuth checks org membership via API (`read:org` scope), stores result on every login. Token refresh rechecks stale (>24h) membership via `GITHUB_SYNC_TOKEN`.
- **Failure policy**: Returning users preserve existing DB org membership on GitHub API failure. New users default to `false`.
- **Seed script**: Sets `visibility='public'` + `submission_source='org_sync'` for org-synced constructs. Validates against enum whitelist.
- **JSON Schema**: Adds `visibility` to `construct.schema.json` (was missing — would reject construct.yaml with visibility declared).
- **Error handling**: Structural DB errors (missing columns) now surface instead of silently returning empty results.

## Review Findings Addressed

| Finding | Severity | Status |
|---------|----------|--------|
| Missing migration file | CRITICAL | Fixed |
| Auth crashes on missing column | CRITICAL | Fixed (by migration) |
| `generateTokens()` never gets `org` | HIGH | Fixed (5 call sites) |
| Seed script ignores visibility | HIGH | Fixed |
| Silent error swallowing | MEDIUM | Fixed (re-throw structural errors) |
| JSON Schema missing visibility | HIGH | Fixed |
| Returning-user failure policy | HIGH | Fixed (preserve DB on API failure) |
| Token refresh staleness recheck | MEDIUM | Fixed (24h recheck via GITHUB_SYNC_TOKEN) |

## Security Audit

15 PASS, 7 ADVISORY, 0 FAIL. No exploitable vulnerabilities. Advisory items addressed:
- Error re-throw wraps in generic message (no schema detail leakage)
- Seed script validates visibility against enum whitelist

## Deployment Sequence

1. Run migration: `psql $DATABASE_URL < apps/api/drizzle/0008_cycle_038_visibility.sql`
2. Deploy API
3. Re-seed if DB was wiped: `DATABASE_URL=... bun tsx scripts/seed-forge-packs.ts`

## Test plan

- [ ] Migration runs idempotently on production DB
- [ ] Registry returns constructs for anonymous users (visibility=public)
- [ ] Auth endpoints return 200 (not 500) for authenticated requests
- [ ] GitHub OAuth login sets `github_org_member` correctly
- [ ] Token refresh preserves org membership
- [ ] Seed script sets visibility=public on all org-synced packs
- [ ] construct.yaml with `visibility: public` passes JSON Schema validation

### DIFF
--- a/.claude/schemas/construct.schema.json
+++ b/.claude/schemas/construct.schema.json
@@ -344,6 +344,12 @@
       "minimum": 0,
       "maximum": 1,
       "description": "Portability score (0.0–1.0) indicating environment independence"
+    },
+    "visibility": {
+      "type": "string",
+      "enum": ["public", "internal", "unlisted"],
+      "default": "internal",
+      "description": "Network visibility — public (listed + anonymous), internal (org-only), unlisted (slug-access only). Cycle-038."
     }
   },
   "additionalProperties": false
--- a/BUTTERFREEZONE.md
+++ b/BUTTERFREEZONE.md
@@ -7,6 +7,19 @@ interfaces:
   core: [/auditing-security, /autonomous-agent, /bridgebuilder-review, /browsing-constructs, /bug-triaging]
   project: [/creating-constructs, /finding-constructs, /linking-constructs, /publishing-constructs, /syncing-constructs]
 dependencies: [git, jq, yq, node]
+ecosystem:
+  - repo: 0xHoneyJar/loa-finn
+    role: runtime
+    interface: hounfour-router
+    protocol: loa-hounfour@8.3.1
+  - repo: 0xHoneyJar/loa-hounfour
+    role: protocol
+    interface: npm-package
+    protocol: loa-hounfour@8.3.1
+  - repo: 0xHoneyJar/arrakis
+    role: distribution
+    interface: jwt-auth
+    protocol: loa-hounfour@8.3.1
 capability_requirements:
   - filesystem: read
   - filesystem: write (scope: state)
@@ -14,7 +27,7 @@ capability_requirements:
   - git: read_write
   - shell: execute
   - github_api: read_write (scope: external)
-version: v2.7.1
+version: v2.9.0
 installation_mode: unknown
 trust_level: L2-verified
 -->
@@ -25,6 +38,37 @@ trust_level: L2-verified
 SaaS platform for distributing, licensing, and monetizing AI agent constructs
 
 The framework provides 35 specialized skills, built with TypeScript/JavaScript, Python, Shell.
+
+## Key Capabilities
+<!-- provenance: DERIVED -->
+The project exposes 15 key entry points across its public API surface.
+
+### .cache/construct-repos/construct-beacon/scripts
+
+- **fail** — Ufail (`./.cache/construct-repos/construct-beacon/scripts/validate.sh:27`)
+- **pass** — Upass (`./.cache/construct-repos/construct-beacon/scripts/validate.sh:26`)
+- **warn** — Uwarn (`./.cache/construct-repos/construct-beacon/scripts/validate.sh:28`)
+
+### .cache/construct-repos/construct-k-hole/scripts/templates
+
+- **DISCOVERY_QUERIES:DiscoveryQuery[]=[** — Udiscovery queries:discovery query[]=[ (`./.cache/construct-repos/construct-k-hole/scripts/templates/research-config.template.ts:28`)
+- **SYNTHESIS_CONTEXT=`Theresearcherwantstobuilddeepexpertisein[** — Usynthesis context=`theresearcherwantstobuilddeepexpertisein[ (`./.cache/construct-repos/construct-k-hole/scripts/templates/research-config.template.ts:75`)
+- **TOPICS:Topic[]=[** — Utopics:topic[]=[ (`./.cache/construct-repos/construct-k-hole/scripts/templates/research-config.template.ts:53`)
+
+### .cache/construct-repos/construct-mibera-codex/.claude/adapters
+
+- **_build_provider_config** — U build provider config (`./.cache/construct-repos/construct-mibera-codex/.claude/adapters/cheval.py:149`)
+- **_error_json** — U error json (`./.cache/construct-repos/construct-mibera-codex/.claude/adapters/cheval.py:74`)
+- **_load_persona** — U load persona (`./.cache/construct-repos/construct-mibera-codex/.claude/adapters/cheval.py:93`)
+- **cmd_invoke** — Ucmd invoke (`./.cache/construct-repos/construct-mibera-codex/.claude/adapters/cheval.py:177`)
+- **cmd_print_config** — Ucmd print config (`./.cache/construct-repos/construct-mibera-codex/.claude/adapters/cheval.py:326`)
+- **cmd_validate_bindings** — Ucmd validate bindings (`./.cache/construct-repos/construct-mibera-codex/.claude/adapters/cheval.py:337`)
+- **main** — Umain (`./.cache/construct-repos/construct-mibera-codex/.claude/adapters/cheval.py:351`)
+
+### .cache/construct-repos/construct-mibera-codex/.claude/adapters/loa_cheval/config
+
+- **LazyValue** — Ulazy value (`./.cache/construct-repos/construct-mibera-codex/.claude/adapters/loa_cheval/config/interpolation.py:41`)
+- **_check_env_allowed** — U check env allowed (`./.cache/construct-repos/construct-mibera-codex/.claude/adapters/loa_cheval/config/interpolation.py:122`)
 
 ## Architecture
 <!-- provenance: DERIVED -->
@@ -85,6 +129,10 @@ Directory structure:
 
 ## Interfaces
 <!-- provenance: DERIVED -->
+### HTTP Routes
+
+- **GET** `/` (`./apps/api/src/app.ts:159`)
+
 ### Skill Commands
 
 #### Loa Core
@@ -116,7 +164,7 @@ Directory structure:
 - **/rtfm-testing** — RTFM Testing Skill
 - **/run-bridge** — Run Bridge — Autonomous Excellence Loop
 - **/run-mode** — Urun mode
-- **/simstim-workflow** — Check post-PR state
+- **/simstim-workflow** — .loa.config.yaml
 - **/translating-for-executives** — Utranslating for executives
 #### Project-Specific
 
@@ -132,21 +180,21 @@ Directory structure:
 | Module | Files | Purpose | Documentation |
 |--------|-------|---------|---------------|
 | `api/` | 3 | API endpoints | \u2014 |
-| `apps/` | 15967 | Uapps | \u2014 |
+| `apps/` | 15747 | Uapps | \u2014 |
 | `audits/` | 1 | Uaudits | \u2014 |
 | `docs/` | 41 | Documentation | \u2014 |
 | `evals/` | 122 | Benchmarking and regression framework for the Loa agent development system. Ensures framework changes don't degrade agent behavior through | [evals/README.md](evals/README.md) |
-| `grimoires/` | 509 | Home to all grimoire directories for the Loa | [grimoires/README.md](grimoires/README.md) |
-| `packages/` | 90 | Upackages | \u2014 |
+| `grimoires/` | 553 | Home to all grimoire directories for the Loa | [grimoires/README.md](grimoires/README.md) |
+| `packages/` | 93 | Upackages | \u2014 |
 | `packs/` | 1 | Upacks | \u2014 |
-| `scripts/` | 27 | Utility scripts | \u2014 |
-| `tests/` | 196 | Test suites | \u2014 |
+| `scripts/` | 30 | Utility scripts | \u2014 |
+| `tests/` | 202 | Test suites | \u2014 |
 
 ## Verification
 <!-- provenance: CODE-FACTUAL -->
 - Trust Level: **L2 — CI Verified**
-- 196 test files across 1 suite
-- CI/CD: GitHub Actions (10 workflows)
+- 202 test files across 1 suite
+- CI/CD: GitHub Actions (12 workflows)
 - Linting: ESLint configured
 - Security: SECURITY.md present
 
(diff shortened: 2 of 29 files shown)


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/loa-constructs#147. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "FIND-001",
      "title": "Redis KEYS command used in production cache invalidation",
      "severity": "HIGH",
      "category": "Performance",
      "file": "apps/api/src/services/redis.ts",
      "description": "The `invalidateConstructCaches` function uses `redis.keys('constructs:list:*')` and `redis.keys('constructs:summary:*')`. The `KEYS` command scans the entire keyspace and is O(N) — Redis documentation explicitly warns against using it in production as it blocks the single-threaded Redis server. Under high key counts this can cause latency spikes or timeouts for all Redis clients.",
      "suggestion": "Use `SCAN`-based iteration (e.g., `scanIterator` or a batched SCAN loop) instead of `KEYS`. Alternatively, maintain a Redis SET of active list/summary cache keys and delete from that set, or use a generation/epoch counter in cache keys that you increment on invalidation (effectively expiring all old keys without scanning).",
      "confidence": 0.95
    },
    {
      "id": "FIND-002",
      "title": "JWT org claim can be stale — overrides fresh DB lookup",
      "severity": "MEDIUM",
      "category": "Security",
      "file": "apps/api/src/middleware/auth.ts",
      "description": "In `requireAuth()` (line ~165) and `optionalAuth()` (line ~210), after fetching the user from DB (which includes `githubOrgMember`), the code overwrites `user.isOrgMember` with `(payload as any).org ?? false` from the JWT claim. JWT access tokens live for 15 minutes, so a user who was removed from the org can still access internal resources until their token expires. The DB value (fetched fresh by `getUserById`) is more authoritative than the cached JWT claim.",
      "suggestion": "Prefer the DB value from `getUserById` over the JWT claim. The JWT `org` claim is useful for performance optimization (skip DB call), but if you're already hitting the DB via `getUserById`, use `user.githubOrgMember` from the DB rather than overriding with the potentially stale JWT claim.",
      "confidence": 0.85
    },
    {
      "id": "FIND-003",
      "title": "Unsafe cast of JWT payload without validation",
      "severity": "MEDIUM",
      "category": "Security",
      "file": "apps/api/src/middleware/auth.ts",
      "description": "The expression `(payload as any).org` at lines ~165 and ~210 uses an unsafe `any` cast to read the `org` claim from the JWT. If a token was issued before this change (no `org` claim), this silently defaults to `false`, which is correct. However, if a malicious or malformed token somehow had `org` set to a truthy non-boolean value, the `?? false` wouldn't catch it (e.g., `org: 1` would be truthy).",
      "suggestion": "Use `typeof (payload as any).org === 'boolean' ? (payload as any).org : false` or add `org` to the `AccessTokenPayload` interface validation (it's already in the interface but the cast bypasses it). Better yet, use `payload.org` directly since the interface already declares it.",
      "confidence": 0.6
    },
    {
      "id": "FIND-004",
      "title": "Migration sets all existing packs to 'public', contradicting the 'internal' default",
      "severity": "MEDIUM",
      "category": "Correctness",
      "file": "apps/api/drizzle/0008_cycle_038_visibility.sql",
      "description": "Line 35: `UPDATE \"packs\" SET \"visibility\" = 'public' WHERE \"visibility\" = 'internal'` updates ALL existing packs to public. The comment explains this is intentional for org-synced packs, but there's a race condition: if any external pack is created between the `ALTER TABLE` (which defaults to 'internal') and this UPDATE, it would also be flipped to 'public'. Additionally, the seed script in `seed-forge-packs.ts` defaults to 'public' for org-synced packs, so there's a potential conflict if the migration and seed run in different orders.",
      "suggestion": "Add a WHERE condition like `WHERE \"submission_source\" = 'org_sync'` or `WHERE \"submission_source\" IS NULL` (since the column also just got added with a default of 'org_sync') to be more precise. This prevents accidental visibility escalation of packs created by concurrent requests.",
      "confidence": 0.7
    },
    {
      "id": "FIND-005",
      "title": "Token refresh org recheck uses server-side token, leaks org membership info via user ID enumeration",
      "severity": "MEDIUM",
      "category": "Security",
      "file": "apps/api/src/routes/auth.ts",
      "description": "The refresh endpoint (line ~232) uses `GITHUB_SYNC_TOKEN` or `GITHUB_TOKEN` to call `https://api.github.com/orgs/${env.CONSTRUCTS_ORG}/members/${ghUser.login}`. This server-side token needs `read:org` scope on the org. If it's a personal access token or GitHub App token with broad access, it can check membership for any user. This is fine functionally, but: (1) `GITHUB_SYNC_TOKEN`/`GITHUB_TOKEN` aren't defined in the env schema (`env.ts`), so they'll be read from `process.env` without validation, and (2) the refresh path makes two sequential GitHub API calls per refresh when stale, adding latency to an auth-critical path.",
      "suggestion": "Add `GITHUB_SYNC_TOKEN` or `GITHUB_TOKEN` to the env schema for validation. Consider making the org recheck async/non-blocking (update DB in background, return current cached value immediately) to avoid adding latency to token refresh.",
      "confidence": 0.8
    },
    {
      "id": "FIND-006",
      "title": "GitHub org membership check uses user's OAuth token instead of server token",
      "severity": "MEDIUM",
      "category": "Security",
      "file": "apps/api/src/routes/oauth.ts",
      "description": "At line ~325, the org membership check during OAuth callback uses `tokenData.access_token` (the user's OAuth token). The `read:org` scope only allows checking public membership or the user's own membership. If the org's membership is private, the check may return 302 (redirect to login) instead of 204, causing `isOrgMember` to be false even for actual members. This works if membership is public, but silently fails for private org membership.",
      "suggestion": "Use a server-side token (like `GITHUB_SYNC_TOKEN`) with org admin/read access for the membership check, or document that the org must have public membership. The refresh path already uses a server token, creating an inconsistency.",
      "confidence": 0.7
    },
    {
      "id": "FIND-007",
      "title": "getVisibilityConditions returns empty array for admin/org, causing invalid SQL",
      "severity": "HIGH",
      "category": "Correctness",
      "file": "apps/api/src/services/constructs.ts",
      "description": "In `getVisibilityConditions()`, when `access.isAdmin || access.isOrgMember`, it returns `[]` (empty array). This is then spread into `and(eq(packs.status, 'published'), ...visConditions)`. While Drizzle's `and()` with just one condition should work, the real issue is in `getConstructsSummary()` at line ~670: `.where(and(eq(packs.status, 'published'), ...visConditions))` — when `visConditions` is empty, `and()` receives only one argument, which should still work in Drizzle. However, if there's ever a case where `and()` receives zero arguments (e.g., conditions array is empty), it would return `undefined`, which drizzle interprets as 'no where clause'. This is fragile.",
      "suggestion": "Instead of returning an empty array for 'no restriction', return an explicit always-true condition like `[sql\\`true\\`]`, or filter out empty arrays before spreading. This makes the intent explicit and prevents future bugs if someone changes the base conditions.",
      "confidence": 0.5
    },
    {
      "id": "FIND-008",
      "title": "Visibility regression test uses fragile string parsing heuristic",
      "severity": "LOW",
      "category": "Maintainability",
      "file": "apps/api/src/routes/visibility-regression.test.ts",
      "description": "The `auditGetPackBySlugCalls` function uses regex and comma-counting heuristics to determine if `getPackBySlug()` calls have a second argument. This will produce false negatives for multi-line calls, calls where the second argument is on the next line, or calls using variable destructuring. It also won't catch calls through re-exported aliases.",
      "suggestion": "Consider using an AST parser (e.g., TypeScript compiler API or `ts-morph`) for more reliable detection, or add a comment-based opt-out mechanism (e.g., `// visibility-checked`) for calls that are intentionally context-free.",
      "confidence": 0.7
    },
    {
      "id": "FIND-009",
      "title": "Status × Submission Source tests only verify object construction, not actual behavior",
      "severity": "LOW",
      "category": "Testing",
      "file": "apps/api/src/services/visibility.test.ts",
      "description": "The '6 Status × Submission Source Combinations' test block (lines ~108–146) and the 'External submission lifecycle' block (lines ~148–181) only construct plain objects and assert their properties match what was set. These tests don't exercise any production code — they're testing JavaScript object literal behavior. For example: `expect(pack.status).toBe('draft')` after `makePack({ status: 'draft' })` is a tautology.",
      "suggestion": "Either test actual service functions (e.g., `createPack` with different statuses, or status transition logic), or remove these tests and document the valid combinations in a type definition or validation schema instead.",
      "confidence": 0.9
    },
    {
      "id": "FIND-010",
      "title": "fetchPackAsConstruct bypasses getPackBySlug — duplicates visibility check",
      "severity": "MEDIUM",
      "category": "Correctness",
      "file": "apps/api/src/services/constructs.ts",
      "description": "The `fetchPackAsConstruct()` function (line ~1033) queries the DB directly and then calls `canAccessPack()` for visibility. However, it does NOT perform the async team ownership fallback that `getPackBySlug()` does (the `isPackOwnerAsync` check). This means team members of a team-owned internal pack will get denied access when the construct detail is fetched via `getConstructBySlug() → fetchPackAsConstruct()` but could access it via `getPackBySlug()` directly.",
      "suggestion": "Either delegate to `getPackBySlug(slug, access)` from `fetchPackAsConstruct`, or replicate the team ownership fallback in `fetchPackAsConstruct`. The current code has two divergent visibility enforcement paths.",
      "confidence": 0.85
    },
    {
      "id": "FIND-011",
      "title": "Admin review endpoint bypasses getPackBySlug visibility — correct design",
      "severity": "PRAISE",
      "category": "Security",
      "file": "apps/api/src/routes/admin.ts",
      "description": "The admin construct review endpoint (PATCH `/constructs/:slug/review`) correctly queries the DB directly instead of going through `getPackBySlug()`, which would filter out `pending_review` packs for non-owners. This is the right approach for admin endpoints that need to access packs in any state.",
      "suggestion": null,
      "confidence": 0.95
    },
    {
      "id": "FIND-012",
      "title": "Comprehensive cache tier segmentation to prevent cache poisoning",
      "severity": "PRAISE",
      "category": "Security",
      "file": "apps/api/src/services/constructs.ts",
      "description": "The decision to segment cache keys by visibility tier (`anon`, `member`, `auth`) and to skip caching entirely for the `auth` tier (to avoid per-user cache pollution) is a well-reasoned security design. The comment explaining the cache poisoning risk for the `auth` tier is particularly valuable.",
      "suggestion": null,
      "confidence": 0.95
    },
    {
      "id": "FIND-013",
      "title": "Graceful org check failure preserves existing DB value",
      "severity": "PRAISE",
      "category": "Resilience",
      "file": "apps/api/src/routes/oauth.ts",
      "description": "The pattern of distinguishing 'API confirmed not-member' from 'API call failed' via `orgCheckSucceeded` and passing `isOrgMember: undefined` on failure (so returning users keep their existing DB value) is a solid resilience pattern that avoids flip-flopping membership status due to transient GitHub API failures.",
      "suggestion": null,
      "confidence": 0.95
    },
    {
      "id": "FIND-014",
      "title": "Webhook sync writes visibility from syncResult without status check",
      "severity": "MEDIUM",
      "category": "Security",
      "file": "apps/api/src/routes/webhooks.ts",
      "description": "At line ~572, the GitHub webhook handler writes `visibility: syncResult.visibility` from the manifest during sync. For external submissions that somehow get past the `pending_review` block (or after approval), this means an external contributor could change their construct.yaml to set `visibility: 'internal'` and potentially hide their public construct from public view, or an org member could change visibility via a git push without admin review.",
      "suggestion": "Consider whether external packs should be allowed to change visibility via manifest sync, or if visibility changes should require explicit admin action for external submissions. At minimum, prevent external packs from setting visibility to 'internal' (which would hide them from public view after being approved as public).",
      "confidence": 0.65
    },
    {
      "id": "FIND-015",
      "title": "Schema error re-throw in fetchPacksAsConstructs is good defensive coding",
      "severity": "PRAISE",
      "category": "Reliability",
      "file": "apps/api/src/services/constructs.ts",
      "description": "The change to re-throw structural DB errors (missing columns/tables) instead of silently returning empty results is an excellent fix. The previous behavior of swallowing all errors made the registry appear empty when migrations were missing, which is much harder to debug than an explicit error.",
      "suggestion": null,
      "confidence": 0.95
    },
    {
      "id": "FIND-016",
      "title": "avatarUrl fallback uses email address instead of null",
      "severity": "LOW",
      "category": "Correctness",
      "file": "apps/api/src/routes/oauth.ts",
      "description": "At line ~161, `avatarUrl: avatarUrl || existingEmail.email` falls back to the user's email address if `avatarUrl` is falsy. This appears to be a pre-existing bug (the diff shows it was `existingEmail.email` before too), but it means the `avatarUrl` field could contain an email address string rather than a URL.",
      "suggestion": "Should likely be `avatarUrl: avatarUrl || undefined` (or keep existing avatar if available). Storing an email address in an `avatarUrl` field will cause rendering issues in any UI that tries to use it as an image source.",
      "confidence": 0.8
    },
    {
      "id": "FIND-017",
      "title": "GitHub user ID stored as bigint but compared as number in TypeScript",
      "severity": "LOW",
      "category": "Correctness",
      "file": "apps/api/src/db/schema.ts",
      "description": "The `githubUserId` column is declared as `bigint('github_user_id', { mode: 'number' })`. GitHub user IDs are currently within safe JavaScript integer range, but as the platform grows, IDs could exceed `Number.MAX_SAFE_INTEGER` (2^53). Using `mode: 'number'` for a bigint column means precision loss for very large IDs.",
      "suggestion": "Consider using `mode: 'bigint'` and handling the BigInt type in TypeScript, or note this as a known limitation that GitHub IDs are currently well within safe integer range (~2 billion active users vs ~9 quadrillion safe integer limit).",
      "confidence": 0.3
    },
    {
      "id": "FIND-018",
      "title": "Missing access context for /reviews GET route (unauthenticated)",
      "severity": "MEDIUM",
      "category": "Security",
      "file": "apps/api/src/routes/packs.ts",
      "description": "The `GET /:slug/reviews` route at line ~1547 calls `getPackBySlug(slug, getAccessContext(c))` but does not use `optionalAuth()` middleware. Without `optionalAuth()`, `getAccessContext(c)` will get `undefined` for `c.get('user')`, meaning all requests are treated as anonymous. Internal pack reviews would not be accessible even to org members via this endpoint. Other similar routes like `GET /:slug/versions` correctly use `optionalAuth()`.",
      "suggestion": "Add `optionalAuth()` middleware to the `GET /:slug/reviews` route so that authenticated users can access reviews for internal packs.",
      "confidence": 0.75
    }
  ]
}
```


## Callouts

PR #152, opened 2026-03-10
feat(cycle-039): Dynamic Labs auth + internal constructs access

zkSoju:
## Summary

- **Dynamic Labs wallet auth** integrated into constructs.network (matching 0xHoneyJar ecosystem pattern)
- **Auth-aware leaderboard**: anonymous visitors see gated CTA, authenticated org members see all 13 internal constructs
- **P0 fix**: `fetchMe` response parsing — `isOrgMember` was always false due to missing envelope unwrap + snake_case mapping
- **P3 fix**: OAuth callback refresh token loss — now stored via HttpOnly cookie route handler

## Sprint Breakdown

| Sprint | Status | Cycles | Files Changed |
|--------|--------|--------|---------------|
| sprint-1 (global sprint-37) | ✅ Complete | 1 | 7 |
| sprint-2 (global sprint-38) | ✅ Complete | 1 | 6 |
| sprint-3 (global sprint-39) | ✅ Complete | 1 | 3 |

**Totals**: 3 sprints, 3 cycles, 16 files changed

### Sprint 1: API Foundation
- Fix `fetchMe` response parsing (P0 — `isOrgMember` always false)
- Add `wallet_address` + `dynamic_user_id` columns (migration `0009`)
- JWKS JWT verification via `jose` (`verify-dynamic-jwt.ts`)
- `POST /v1/auth/dynamic` endpoint — Dynamic Labs JWT → our API JWT
- OAuth refresh token fix (HttpOnly cookie route handler)

### Sprint 2: Explorer Auth Integration
- Dynamic Labs provider setup (`DynamicContextProvider > WagmiProvider > QueryClientProvider > DynamicWagmiConnector`)
- Lazy-loaded via `next/dynamic` with `ssr: false`
- `connectDynamic()` action in auth store
- `DynamicConnectButton` component with wallet state management
- `AuthNav` header island with hydration safety

### Sprint 3: Auth-Aware Public Pages
- `AuthAwareConstructList` — render prop pattern for ISR + client auth overlay
- Homepage leaderboard wrapped with auth-aware CTA
- Catalog page wrapped with auth-aware CTA
- Internal construct badges for authenticated views

## Env Vars Required

| Variable | Where | Value |
|----------|-------|-------|
| `DYNAMIC_ENVIRONMENT_ID` | Railway (API) | From Dynamic Labs dashboard |
| `NEXT_PUBLIC_DYNAMIC_ENVIRONMENT_ID` | Vercel (Explorer) | Same value |

## DB Migration Required

```sql
-- apps/api/drizzle/0009_cycle_039_dynamic_auth.sql
ALTER TABLE "users" ADD COLUMN IF NOT EXISTS "wallet_address" varchar(42);
ALTER TABLE "users" ADD COLUMN IF NOT EXISTS "dynamic_user_id" varchar(100);
CREATE UNIQUE INDEX IF NOT EXISTS "idx_users_wallet" ON "users" ("wallet_address") WHERE "wallet_address" IS NOT NULL;
CREATE UNIQUE INDEX IF NOT EXISTS "idx_users_dynamic" ON "users" ("dynamic_user_id") WHERE "dynamic_user_id" IS NOT NULL;
```

## Test plan

- [ ] Apply migration `0009` to production Supabase
- [ ] Set `DYNAMIC_ENVIRONMENT_ID` on Railway, `NEXT_PUBLIC_DYNAMIC_ENVIRONMENT_ID` on Vercel
- [ ] Visit constructs.network incognito → "Connect to explore" CTA, 0 constructs
- [ ] Connect wallet → Dynamic Labs modal → authenticate
- [ ] Wallet only → "Link GitHub for internal access" prompt
- [ ] Link GitHub (org member) → all 13 constructs appear
- [ ] Refresh page → session persists
- [ ] Existing OAuth login still works alongside Dynamic Labs

🤖 Generated autonomously with Run Mode

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/.claude/schemas/construct.schema.json
+++ b/.claude/schemas/construct.schema.json
@@ -344,6 +344,12 @@
       "minimum": 0,
       "maximum": 1,
       "description": "Portability score (0.0–1.0) indicating environment independence"
+    },
+    "visibility": {
+      "type": "string",
+      "enum": ["public", "internal", "unlisted"],
+      "default": "internal",
+      "description": "Network visibility — public (listed + anonymous), internal (org-only), unlisted (slug-access only). Cycle-038."
     }
   },
   "additionalProperties": false
--- a/BUTTERFREEZONE.md
+++ b/BUTTERFREEZONE.md
@@ -7,6 +7,19 @@ interfaces:
   core: [/auditing-security, /autonomous-agent, /bridgebuilder-review, /browsing-constructs, /bug-triaging]
   project: [/creating-constructs, /finding-constructs, /linking-constructs, /publishing-constructs, /syncing-constructs]
 dependencies: [git, jq, yq, node]
+ecosystem:
+  - repo: 0xHoneyJar/loa-finn
+    role: runtime
+    interface: hounfour-router
+    protocol: loa-hounfour@8.3.1
+  - repo: 0xHoneyJar/loa-hounfour
+    role: protocol
+    interface: npm-package
+    protocol: loa-hounfour@8.3.1
+  - repo: 0xHoneyJar/arrakis
+    role: distribution
+    interface: jwt-auth
+    protocol: loa-hounfour@8.3.1
 capability_requirements:
   - filesystem: read
   - filesystem: write (scope: state)
@@ -14,7 +27,7 @@ capability_requirements:
   - git: read_write
   - shell: execute
   - github_api: read_write (scope: external)
-version: v2.7.1
+version: v2.9.0
 installation_mode: unknown
 trust_level: L2-verified
 -->
@@ -25,6 +38,37 @@ trust_level: L2-verified
 SaaS platform for distributing, licensing, and monetizing AI agent constructs
 
 The framework provides 35 specialized skills, built with TypeScript/JavaScript, Python, Shell.
+
+## Key Capabilities
+<!-- provenance: DERIVED -->
+The project exposes 15 key entry points across its public API surface.
+
+### .cache/construct-repos/construct-beacon/scripts
+
+- **fail** — Ufail (`./.cache/construct-repos/construct-beacon/scripts/validate.sh:27`)
+- **pass** — Upass (`./.cache/construct-repos/construct-beacon/scripts/validate.sh:26`)
+- **warn** — Uwarn (`./.cache/construct-repos/construct-beacon/scripts/validate.sh:28`)
+
+### .cache/construct-repos/construct-k-hole/scripts/templates
+
+- **DISCOVERY_QUERIES:DiscoveryQuery[]=[** — Udiscovery queries:discovery query[]=[ (`./.cache/construct-repos/construct-k-hole/scripts/templates/research-config.template.ts:28`)
+- **SYNTHESIS_CONTEXT=`Theresearcherwantstobuilddeepexpertisein[** — Usynthesis context=`theresearcherwantstobuilddeepexpertisein[ (`./.cache/construct-repos/construct-k-hole/scripts/templates/research-config.template.ts:75`)
+- **TOPICS:Topic[]=[** — Utopics:topic[]=[ (`./.cache/construct-repos/construct-k-hole/scripts/templates/research-config.template.ts:53`)
+
+### .cache/construct-repos/construct-mibera-codex/.claude/adapters
+
+- **_build_provider_config** — U build provider config (`./.cache/construct-repos/construct-mibera-codex/.claude/adapters/cheval.py:149`)
+- **_error_json** — U error json (`./.cache/construct-repos/construct-mibera-codex/.claude/adapters/cheval.py:74`)
+- **_load_persona** — U load persona (`./.cache/construct-repos/construct-mibera-codex/.claude/adapters/cheval.py:93`)
+- **cmd_invoke** — Ucmd invoke (`./.cache/construct-repos/construct-mibera-codex/.claude/adapters/cheval.py:177`)
+- **cmd_print_config** — Ucmd print config (`./.cache/construct-repos/construct-mibera-codex/.claude/adapters/cheval.py:326`)
+- **cmd_validate_bindings** — Ucmd validate bindings (`./.cache/construct-repos/construct-mibera-codex/.claude/adapters/cheval.py:337`)
+- **main** — Umain (`./.cache/construct-repos/construct-mibera-codex/.claude/adapters/cheval.py:351`)
+
+### .cache/construct-repos/construct-mibera-codex/.claude/adapters/loa_cheval/config
+
+- **LazyValue** — Ulazy value (`./.cache/construct-repos/construct-mibera-codex/.claude/adapters/loa_cheval/config/interpolation.py:41`)
+- **_check_env_allowed** — U check env allowed (`./.cache/construct-repos/construct-mibera-codex/.claude/adapters/loa_cheval/config/interpolation.py:122`)
 
 ## Architecture
 <!-- provenance: DERIVED -->
@@ -85,6 +129,10 @@ Directory structure:
 
 ## Interfaces
 <!-- provenance: DERIVED -->
+### HTTP Routes
+
+- **GET** `/` (`./apps/api/src/app.ts:159`)
+
 ### Skill Commands
 
 #### Loa Core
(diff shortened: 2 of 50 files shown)


zkSoju (commented):
# Bridgebuilder Review — PR #152

## Summary

This is a substantial, well-structured PR that stacks cycle-039 (Dynamic Labs wallet auth + auth-aware pages) on top of cycle-038 (visibility system). The `fetchMe` P0 fix, JWKS verification module, and auth-aware component architecture are solid. The primary concerns are around the CSRF validation in the refresh token route handler (origin-check bypass possible in dev), a potential race condition in the Dynamic connect button, and the use of `process.env` inside a module-level singleton for JWKS setup. Overall, this is above-average work with strong attention to the fail-stale org membership pattern and ecosystem consistency.

---

## Findings

### FIND-001 [Security] — Origin check bypass in set-refresh route handler — `high`

**File**: `apps/explorer/app/api/auth/set-refresh/route.ts`, lines ~14-17

The CSRF origin check allows any `*.vercel.app` subdomain:

```typescript
if (process.env.NODE_ENV === 'production' && origin && !allowedOrigins.includes(origin) && !origin.endsWith('.vercel.app'))
```

Any attacker who deploys a Vercel project can forge requests from `evil.vercel.app` to this endpoint. The `.endsWith('.vercel.app')` pattern is too broad.

**FAANG Parallel**: This is the exact class of origin-check bypass that the Chrome security team documented in their CORS relaxation audit — wildcard subdomain allowlists undermine the security model they're supposed to enforce.

**Metaphor**: This is like a bouncer who lets anyone in wearing a Vercel t-shirt. You want to check the guest list, not the brand of clothing.

**Recommendation**: Restrict to your specific Vercel deployment URLs. Use a regex like `/^https:\/\/loa-constructs(-[a-z0-9]+)?\.vercel\.app$/` or maintain an explicit allowlist. Alternatively, use a CSRF token in a cookie-to-header pattern.

---

### FIND-002 [Quality] — Race condition in Dynamic connect button `useEffect` — `medium`

**File**: `apps/explorer/components/auth/dynamic-connect-button.tsx`, lines ~44-47

```typescript
useEffect(() => {
  if (primaryWallet && !isAuthenticated && !isExchanging) {
    handleAuthSuccess();
  }
}, [primaryWallet, isAuthenticated, isExchanging, handleAuthSuccess]);
```

If `primaryWallet` changes rapidly (e.g., user switches wallets in the Dynamic modal), multiple `handleAuthSuccess()` calls can fire concurrently. The `isExchanging` state update is async, so there is a window where the guard fails and two exchange requests happen simultaneously. The second will either fail (duplicate user creation) or succeed with stale state.

**Recommendation**: Use a ref-based lock (`const exchangingRef = useRef(false)`) alongside the state boolean. Check the ref synchronously at the top of `handleAuthSuccess`.

---

### FIND-003 [Security] — JWKS singleton uses `process.env` at call time, not startup — `medium`

**File**: `apps/api/src/lib/verify-dynamic-jwt.ts`, lines ~26-33

```typescript
function getJWKS(): ReturnType<typeof createRemoteJWKSet> {
  if (!jwks) {
    const envId = process.env.DYNAMIC_ENVIRONMENT_ID;
    if (!envId) throw new Error('DYNAMIC_ENVIRONMENT_ID environment variable is required');
```

The lazy singleton reads `process.env.DYNAMIC_ENVIRONMENT_ID` at first invocation, not at startup. If the env var is missing, the first Dynamic auth request gets a 500 error with a stack trace that reveals the env var name. This is a minor information leak.

**Recommendation**: Validate `DYNAMIC_ENVIRONMENT_ID` at startup alongside other env vars in `config/env.ts` (where the Zod schema already validates it as optional). The schema already has `DYNAMIC_ENVIRONMENT_ID: z.string().optional()` — consider making it required when the Dynamic auth route is mounted, or add an early validation in the route file itself.

---

### FIND-004 [Quality] — `fetchMe` type casting without runtime validation — `medium`

**File**: `apps/explorer/lib/api/auth.ts`, lines ~56-66

```typescript
const u = response.user;
return {
  id: u.id as string,
  email: u.email as string,
  name: (u.name as string) ?? null,
  role: (u.tier as string) ?? 'free',
```

The response is cast with `as` without runtime validation. If the API shape changes or the envelope structure evolves, this silently produces incorrect data rather than throwing.

**Decision Trail**: Document that the API `/auth/me` response contract is `{ user: { id, email, name, tier, email_verified, is_org_member, created_at } }` in a shared types file or as a Zod schema. This prevents future drift between API and client.

**Recommendation**: Add a lightweight Zod parse or at minimum a runtime check that `response.user` exists and `u.id` is truthy before mapping.

---

### FIND-005 [Operational] — Missing rate limiter on `/v1/auth/dynamic` endpoint — `medium`

**File**: `apps/api/src/routes/dynamic-auth.ts`

The PRD and SDD both specify "Rate-limited to 10 req/min per IP" on the Dynamic auth endpoint. However, the route file does not apply `authRateLimiter` middleware. The route is mounted directly: `dynamicAuth.post('/', async (c) => { ... })` with no rate limiter.

**FAANG Parallel**: Auth endpoints without rate limiting are the #1 target for credential stuffing. Cloudflare's 2024 threat report showed auth endpoints receive 3x the attack traffic of any other endpoint class.

**Recommendation**: Add `authRateLimiter` middleware to the Dynamic auth route, same as the existing `/auth/login` and `/auth/register` routes.

---

### FIND-006 [Quality] — `resolveUser` OR query may match wrong user — `high`

**File**: `apps/api/src/routes/dynamic-auth.ts`, lines ~83-104

The `resolveUser` function builds an OR query across `dynamic_user_id`, `wallet_address`, and `github_username`. If user A has wallet `0xABC` and user B has github username `alice`, and a new Dynamic JWT comes in with wallet `0xABC` and github `alice`, the OR query matches either user nondeterministically (`.limit(1)` with no ORDER BY). The matched user gets the other user's identity fields overwritten.

**Metaphor**: Imagine a hotel reception desk that looks up guests by "name OR passport number OR loyalty card." If two different guests share a loyalty card number, one of them gets the other's room key.

**Recommendation**: Prioritize lookups with sequential queries instead of a single OR:
1. First, try `dynamic_user_id` (exact match, most specific)
2. If not found, try `wallet_address`
3. If not found, try `github_username`
This eliminates the ambiguity of OR-matching across identity types. The performance cost of 2-3 sequential queries on indexed unique columns is negligible.

---

### FIND-007 [Test Coverage] — No unit or integration tests for Dynamic auth flow — `medium`

The PR adds a complete new auth provider (Dynamic Labs) with JWT verification, user resolution, org membership checks, and token exchange — but ships zero tests. The `verify-dynamic-jwt.ts` module, `resolveUser` function, and `connectDynamic` store action are all untested.

**Decision Trail**: Document the testing strategy. At minimum: (1) unit test `verifyDynamicJWT` with a mock JWKS, (2) unit test `resolveUser` with the three priority paths, (3) integration test for the full exchange flow. The existing `__tests__/` pattern in the codebase shows the team values testing.

---

## Positive Callouts

### [Security] DB-authoritative org membership (FIND-002 fix from cycle-038) — `praise`

The decision to make `getUserById()` DB values authoritative over JWT `org` claims is excellent. This eliminates an entire class of privilege escalation where stale JWT claims grant access after org removal. The comment trail explaining the reasoning is exemplary documentation.

### [Quality] Graceful DynamicProvider fallback — `praise`

```typescript
if (!environmentId) {
  return <>{children}</>;
}
```

The provider silently degrades when `NEXT_PUBLIC_DYNAMIC_ENVIRONMENT_ID` is not set. This means the app works without Dynamic Labs configured — essential for local development and CI. This is the kind of defensive coding that prevents "works on my machine" failures.

### [Operational] OAuth org check distinguishes authoritative vs transient failures — `praise`

The `orgCheckSucceeded` boolean in `oauth.ts` that distinguishes "API confirmed not-member" from "API call failed" is a pattern worth reusing. The fail-stale policy (preserve existing DB value on transient errors) is the right tradeoff for a 24h-rechecked cache.

### [Quality] Migration idempotency — `praise`

Both migration files (`0008`, `0009`) use `IF NOT EXISTS` and `DO $$ BEGIN ... EXCEPTION WHEN duplicate_object` blocks. This makes them safe to re-run, which is critical for the Railway deployment pattern where migrations may be retried.

---

## Verdict

**REQUEST_CHANGES** — FIND-001 (origin bypass) and FIND-006 (OR query ambiguity) should be addressed before merge. The remaining findings are improvements that can be tracked as follow-ups.

PR #154, opened 2026-03-10
feat: internal dashboard + Convex real-time integration (cycle-040)

zkSoju:
## Summary

- **Internal dashboard** with auth-gated routing, sidebar navigation, and admin section
- **API key self-service** — full CRUD with 10-key limit, prefix-only listing, ownership-checked revocation
- **Convex real-time layer** — hot/cold architecture replicating midi-interface pattern with graceful degradation (fully functional without Convex env vars)

## Sprint Breakdown

| Sprint | Description | Cycles | Files Changed |
|--------|-------------|--------|---------------|
| Sprint 1 | `is_admin` pipeline: DB → API → auth store | 1 | 7 |
| Sprint 2 | Dashboard shell: layout, sidebar, pages | 1 | 6 |
| Sprint 3 | API key management: CRUD endpoints + UI | 1 | 6 |
| Sprint 4 | Convex foundation: schema, provider, server client, webhook | 1 | 11 |
| Sprint 5 | Live dashboard: analytics, install feed, reconciliation cron | 1 | 6 |

**Totals:** 5 sprints, 5 cycles, 36 files changed, 1508 insertions

## Architecture

- **Hot/Cold**: Convex WebSocket reads (real-time), Supabase PostgreSQL writes (source of truth)
- **Graceful Degradation**: `ConvexProvider` is no-op without `NEXT_PUBLIC_CONVEX_URL`
- **Two-Layer Auth**: Next.js middleware (cookie soft gate) + client-side auth store (hard gate)
- **Write-Key Gating**: Server-to-server Convex mutations require `CONVEX_WRITE_KEY` shared secret
- **Fire-and-Forget Webhook**: API install tracking → explorer BFF → Convex (non-blocking)

## Setup Required

Before deployment:

1. **Install Convex dependency**: `cd apps/explorer && bun add convex`
2. **Convex Cloud project**: Create project, run `npx convex deploy`
3. **Run migration**: `0010_admin_wallet.sql` on production Supabase
4. **Environment variables**:
   - `NEXT_PUBLIC_CONVEX_URL` — Convex deployment URL
   - `CONVEX_URL` — Server-side Convex URL
   - `CONVEX_WRITE_KEY` — Shared secret for mutations
   - `CONVEX_WEBHOOK_URL` — Explorer BFF webhook endpoint
   - `CRON_SECRET` — Vercel cron auth token

## Test plan

- [ ] Verify dashboard routes redirect unauthenticated users to `/?login=required`
- [ ] Verify `isAdmin` gates admin sections (analytics, admin sidebar items)
- [ ] Create / list / revoke API keys through dashboard UI
- [ ] Confirm app builds and runs without any Convex env vars (graceful degradation)
- [ ] With Convex configured: verify live install feed updates via WebSocket
- [ ] Verify reconciliation cron returns 401 without valid `CRON_SECRET`
- [ ] Run migration on staging and confirm `is_admin` column works

### DIFF
--- a/apps/api/src/app.ts
+++ b/apps/api/src/app.ts
@@ -23,6 +23,7 @@ import { signalsRouter } from './routes/signals.js';
 import { constructsRouter } from './routes/constructs.js';
 import { categoriesRouter } from './routes/categories.js';
 import { adminRouter } from './routes/admin.js';
+import { keysRouter } from './routes/keys.js';
 import { publicKeysRouter } from './routes/public-keys.js';
 import { apiRateLimiter } from './middleware/rate-limiter.js';
 import { env } from './config/env.js';
@@ -143,6 +144,9 @@ v1.route('/constructs', constructsRouter);
 
 // Categories routes (taxonomy)
 v1.route('/categories', categoriesRouter);
+
+// API key management (cycle-040)
+v1.route('/keys', keysRouter);
 
 // Admin routes (requires admin role)
 v1.route('/admin', adminRouter);
--- a/apps/api/src/db/migrations/0010_admin_wallet.sql
+++ b/apps/api/src/db/migrations/0010_admin_wallet.sql
@@ -1,0 +1,4 @@
+-- cycle-040: Promote primary admin wallet
+-- Idempotent — SET is_admin = true is a no-op if already true
+UPDATE users SET is_admin = true
+WHERE LOWER(wallet_address) = LOWER('0x79092A805f1cf9B0F5bE3c5A296De6e51c1DEd34');
--- a/apps/api/src/middleware/auth.ts
+++ b/apps/api/src/middleware/auth.ts
@@ -21,6 +21,7 @@ export interface AuthUser {
   tier: 'free' | 'pro' | 'team' | 'enterprise';
   role?: 'user' | 'admin' | 'super_admin';
   isOrgMember: boolean; // cycle-038: GitHub org membership
+  walletAddress: string | null; // cycle-040: dashboard
 }
 
 // Type augmentation for Hono context
@@ -68,6 +69,7 @@ async function getUserById(userId: string): Promise<AuthUser | null> {
       emailVerified: users.emailVerified,
       isAdmin: users.isAdmin,
       githubOrgMember: users.githubOrgMember, // cycle-038
+      walletAddress: users.walletAddress, // cycle-040
     })
     .from(users)
     .where(eq(users.id, userId))
@@ -88,6 +90,7 @@ async function getUserById(userId: string): Promise<AuthUser | null> {
     tier: effectiveTier.tier,
     role: user.isAdmin ? 'admin' : 'user',
     isOrgMember: user.githubOrgMember ?? false, // cycle-038
+    walletAddress: user.walletAddress ?? null, // cycle-040
   };
 }
 
--- a/apps/api/src/routes/admin.ts
+++ b/apps/api/src/routes/admin.ts
@@ -1012,3 +1012,41 @@ adminRouter.get('/analytics/packs', zValidator('query', analyticsQuerySchema), a
     recent_packs: recentPacks,
   });
 });
+
+/**
+ * GET /v1/admin/analytics/installs
+ * Daily bucketed install counts — cycle-040
+ */
+const installQuerySchema = z.object({
+  pack_id: z.string().uuid().optional(),
+  period: z.enum(['7d', '30d', '90d']).default('30d'),
+});
+
+adminRouter.get('/analytics/installs', zValidator('query', installQuerySchema), async (c) => {
+  const { pack_id: packId, period } = c.req.valid('query');
+
+  const days = period === '7d' ? 7 : period === '90d' ? 90 : 30;
+  const since = new Date();
+  since.setDate(since.getDate() - days);
+
+  const conditions = [gte(packInstallations.createdAt, since)];
+  if (packId) {
+    conditions.push(eq(packInstallations.packId, packId));
+  }
+
+  const dailyCounts = await db
+    .select({
+      date: sql<string>`date_trunc('day', ${packInstallations.createdAt})::date::text`,
+      count: count(),
+    })
+    .from(packInstallations)
+    .where(and(...conditions))
+    .groupBy(sql`date_trunc('day', ${packInstallations.createdAt})`)
+    .orderBy(sql`date_trunc('day', ${packInstallations.createdAt})`);
+
+  return c.json({
+    period,
+    pack_id: packId ?? null,
+    buckets: dailyCounts,
+  });
+});
--- a/apps/api/src/routes/auth.ts
+++ b/apps/api/src/routes/auth.ts
@@ -475,6 +475,8 @@ auth.get('/me', requireAuth(), async (c) => {
       name: user.name,
       email_verified: user.emailVerified,
       is_org_member: user.isOrgMember,
+      is_admin: user.role === 'admin',
+      wallet_address: user.walletAddress ?? null,
       tier: user.tier,
     },
   });
--- a/apps/api/src/routes/keys.ts
+++ b/apps/api/src/routes/keys.ts
@@ -33,6 +33,16 @@ keys.post('/', requireAuth(), async (c) => {
     return c.json({ error: 'Maximum 10 active API keys per user' }, 400);
   }
 
+  const ALLOWED_SCOPES = ['read:skills', 'write:installs', 'read:analytics'];
+  const scopes = (body.scopes || ['read:skills', 'write:installs']).filter(
+    (s: string) => ALLOWED_SCOPES.includes(s),
+  );
+  if (scopes.length === 0) {
+    return c.json({ error: 'No valid scopes provided' }, 400);
+  }
+
+  const name = (body.name || 'Unnamed key').slice(0, 100).trim() || 'Unnamed key';
+
   const { key, prefix } = generateApiKey();
   const hash = await hashApiKey(key);
 
@@ -42,8 +52,8 @@ keys.post('/', requireAuth(), async (c) => {
       userId,
       keyPrefix: prefix,
       keyHash: hash,
-      name: body.name || 'Unnamed key',
-      scopes: body.scopes || ['read:skills', 'write:installs'],
+      name,
+      scopes,
     })
     .returning();
 
--- a/apps/api/src/routes/keys.ts
+++ b/apps/api/src/routes/keys.ts
@@ -1,0 +1,101 @@
+import { Hono } from 'hono';
+import { requireAuth } from '../middleware/auth.js';
+import { generateApiKey, hashApiKey } from '../services/auth.js';
+import { apiKeys } from '../db/schema.js';
+import { db } from '../db/index.js';
+import { eq, and, count } from 'drizzle-orm';
+
+/**
+ * API Key Management — cycle-040
+ * CRUD endpoints for self-service API key management.
+ * Uses existing generateApiKey/hashApiKey from services/auth.ts.
+ */
+const keys = new Hono();
+
+// POST /v1/keys — Create API key
+keys.post('/', requireAuth(), async (c) => {
+  const userId = c.get('userId');
+
+  let body: { name?: string; scopes?: string[] };
+  try {
+    body = await c.req.json();
+  } catch {
+    body = {};
+  }
+
+  // Enforce 10-key limit
+  const [{ value: activeCount }] = await db
+    .select({ value: count() })
+    .from(apiKeys)
+    .where(and(eq(apiKeys.userId, userId), eq(apiKeys.revoked, false)));
+
+  if (activeCount >= 10) {
+    return c.json({ error: 'Maximum 10 active API keys per user' }, 400);
+  }
+
+  const { key, prefix } = generateApiKey();
+  const hash = await hashApiKey(key);
+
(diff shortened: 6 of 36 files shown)


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/loa-constructs#154. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "CRIT-001",
      "title": "Admin dashboard link exposed to all authenticated users",
      "severity": "HIGH",
      "category": "authorization",
      "file": "apps/explorer/components/layout/auth-nav.tsx:30-36",
      "description": "The Dashboard link is shown to all authenticated users (`isAuthenticated`), not just org members or admins. While the dashboard layout has a client-side auth guard, the admin analytics section inside the dashboard overview page relies purely on the client-side `isAdmin` flag from the auth store. The `/v1/admin/analytics` endpoint presumably has server-side admin checks, but the new `/v1/admin/analytics/installs` endpoint is mounted on `adminRouter` — need to verify it inherits admin middleware. The `fetchDashboard` helper hits `/constructs?per_page=100` for all users which may not be an issue, but the pattern of trusting client-side `isAdmin` for conditional rendering of admin sections is fragile.",
      "suggestion": "Verify that all admin routes have server-side authorization. The client-side `isAdmin` flag should only control UI rendering, never be the sole authorization gate. Consider whether the Dashboard link should be gated on `isOrgMember` or `isAdmin` rather than just `isAuthenticated`.",
      "confidence": 0.7
    },
    {
      "id": "CRIT-002",
      "title": "Middleware auth check uses only cookie presence, not validity",
      "severity": "HIGH",
      "category": "security",
      "file": "apps/explorer/middleware.ts:5-8",
      "description": "The Next.js middleware only checks for the existence of the `access_token` cookie, not whether it's valid, expired, or well-formed. Any cookie value (including an empty string or garbage) will pass this check. An attacker could set `access_token=x` to bypass the redirect.",
      "suggestion": "At minimum, check that the token is non-empty and has a plausible JWT structure (three dot-separated parts). For stronger protection, decode and verify the expiry claim in edge middleware, or validate against the API.",
      "confidence": 0.9
    },
    {
      "id": "CRIT-003",
      "title": "No scope validation on API key creation",
      "severity": "HIGH",
      "category": "security",
      "file": "apps/api/src/routes/keys.ts:39-42",
      "description": "The `POST /v1/keys` endpoint accepts arbitrary scope strings from the user (`body.scopes || ['read:skills', 'write:installs']`). There is no validation that the provided scopes are from an allowed set. A user could submit `['admin:all', 'write:users']` or any string, potentially granting themselves elevated privileges if scope-based authorization is later enforced.",
      "suggestion": "Add a whitelist of allowed scopes and validate `body.scopes` against it before inserting into the database. Return a 400 error for invalid scopes.",
      "confidence": 0.95
    },
    {
      "id": "CRIT-004",
      "title": "Hardcoded admin promotion in migration file",
      "severity": "MEDIUM",
      "category": "security",
      "file": "apps/api/src/db/migrations/0010_admin_wallet.sql:3-4",
      "description": "A specific wallet address is hardcoded as admin in a migration file. While idempotent, this permanently embeds a privileged address in source control. If this address is compromised or ownership changes, a new migration would be needed. Additionally, the address is publicly visible in the repository.",
      "suggestion": "Consider using a seed script or environment variable-based admin bootstrapping rather than hardcoding addresses in migrations. If this approach is intentional, document the process for revoking/rotating admin access.",
      "confidence": 0.6
    },
    {
      "id": "HIGH-001",
      "title": "Convex write key passed through client-accessible mutation args",
      "severity": "HIGH",
      "category": "security",
      "file": "apps/explorer/convex/installEvents.ts:16-32",
      "description": "The `record` mutation accepts `writeKey` as a plain argument and compares it against `process.env.CONVEX_WRITE_KEY`. However, Convex mutations are callable by any client with the Convex deployment URL. The `CONVEX_WRITE_KEY` is also passed from the Next.js API route (`apps/explorer/app/api/convex/install/route.ts:21`) which correctly authenticates the inbound webhook, but then forwards the key into the Convex mutation args. If anyone discovers the Convex deployment URL, they can call `record` directly — the write key check provides some protection but the key could be extracted from network traffic if the BFF route is called.",
      "suggestion": "Use Convex's `internalMutation` for `record` so it cannot be called from external clients at all. Have the Next.js API route use `ConvexHttpClient.mutation` with an internal function reference instead.",
      "confidence": 0.85
    },
    {
      "id": "HIGH-002",
      "title": "Same vulnerability in syncStatus.upsert — should be internal mutation",
      "severity": "HIGH",
      "category": "security",
      "file": "apps/explorer/convex/syncStatus.ts:20-50",
      "description": "The `syncStatus.upsert` mutation uses the same pattern of accepting a `writeKey` argument for authorization but is exported as a public `mutation`. Any client with the Convex URL can call this and attempt to brute-force or guess the write key.",
      "suggestion": "Convert to `internalMutation` and only call from server-side Convex functions or the HTTP client.",
      "confidence": 0.85
    },
    {
      "id": "MED-001",
      "title": "Reconcile cron endpoint performs no meaningful work",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "apps/explorer/app/api/cron/reconcile/route.ts:56-64",
      "description": "The reconciliation cron handler has a placeholder comment stating the implementation requires an install events API. The while loop immediately breaks, the fetched API response is never used, and `reconciled` is always 0. This endpoint runs every 15 minutes (per vercel.json) consuming resources for no effect, and the unused `since` variable on line 47 confirms incomplete implementation.",
      "suggestion": "Either implement the reconciliation logic or remove the cron job until it's ready. Running a no-op every 15 minutes wastes Vercel function invocations and could confuse monitoring.",
      "confidence": 0.95
    },
    {
      "id": "MED-002",
      "title": "Conditional hook call violates Rules of Hooks",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "apps/explorer/hooks/use-dashboard-presence.ts:12-17",
      "description": "The `useDashboardPresence` function contains an early return before calling `usePresenceInner()`, which itself contains hooks (`useMutation`, `useQuery`, `useAuthStore`, `useEffect`). The eslint-disable comment on line 15 acknowledges this violation. While the `CONVEX_AVAILABLE` constant won't change at runtime (it's based on an env var), this pattern can cause subtle issues with React's hook ordering expectations and will confuse static analysis tools.",
      "suggestion": "Restructure to always call hooks but make them no-ops when Convex is unavailable. For example, create a single hook that checks the flag internally after all hooks are called, or use a provider-level feature flag.",
      "confidence": 0.7
    },
    {
      "id": "MED-003",
      "title": "Same conditional hook pattern in LiveInstallFeed",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "apps/explorer/components/dashboard/live-install-feed.tsx:18-23",
      "description": "The `LiveInstallFeed` component returns `<FeedOffline />` before `<LiveFeedInner />` is rendered. While this is technically safe (it's component composition, not conditional hooks within the same component), the `CONVEX_AVAILABLE` check at module level using `process.env.NEXT_PUBLIC_CONVEX_URL` is evaluated at build time in Next.js, not at runtime. If the env var is set after the build, this check will be stale.",
      "suggestion": "Document that `NEXT_PUBLIC_CONVEX_URL` must be available at build time for this feature flag to work, or use a runtime check.",
      "confidence": 0.6
    },
    {
      "id": "MED-004",
      "title": "Dashboard presence cleanup may not scale efficiently",
      "severity": "MEDIUM",
      "category": "performance",
      "file": "apps/explorer/convex/dashboardPresence.ts:44-53",
      "description": "The `cleanupExpired` internal mutation uses `.collect()` which loads all expired documents into memory before deleting them one by one. Running every 30 seconds (per crons.ts), this could become expensive if many presence records accumulate. Convex charges per document read and write.",
      "suggestion": "Add a `.take(100)` limit to the cleanup query to bound work per invocation. The 30-second interval will catch any remaining documents in subsequent runs.",
      "confidence": 0.5
    },
    {
      "id": "MED-005",
      "title": "listOnline query scans all presence records",
      "severity": "MEDIUM",
      "category": "performance",
      "file": "apps/explorer/convex/dashboardPresence.ts:34-41",
      "description": "The `listOnline` query uses `.collect()` on the `by_expires` index without any range filter, then filters in memory with `.filter((p) => p.expiresAt > now)`. This reads all presence records (including soon-to-be-cleaned-up expired ones) on every query invocation.",
      "suggestion": "Use the index range: `.withIndex('by_expires', (q) => q.gt('expiresAt', now))` to let Convex filter at the storage layer instead of in memory.",
      "confidence": 0.85
    },
    {
      "id": "MED-006",
      "title": "No input validation on key name length",
      "severity": "LOW",
      "category": "security",
      "file": "apps/api/src/routes/keys.ts:40",
      "description": "The API key name is set from user input with only a fallback (`body.name || 'Unnamed key'`) but no length or content validation. A user could submit an extremely long name or include malicious content that could cause issues in database storage or UI rendering.",
      "suggestion": "Add validation: trim, enforce max length (e.g., 100 chars), and sanitize the name string.",
      "confidence": 0.7
    },
    {
      "id": "LOW-001",
      "title": "Convex server client caches null when URL is not set",
      "severity": "LOW",
      "category": "correctness",
      "file": "apps/explorer/lib/convex/server.ts:5-9",
      "description": "The `getConvexClient` function caches the result in a module-level variable, including `null`. If the function is first called before the environment variable is available (e.g., during module initialization in some edge cases), it will permanently return `null` for the lifetime of the process even if the env var is later set.",
      "suggestion": "Only cache non-null results, or check the env var on every call since the object creation is lightweight.",
      "confidence": 0.5
    },
    {
      "id": "LOW-002",
      "title": "Missing body validation in convex install webhook route",
      "severity": "MEDIUM",
      "category": "security",
      "file": "apps/explorer/app/api/convex/install/route.ts:19-26",
      "description": "The POST handler calls `req.json()` without a try/catch and passes body properties directly to the Convex mutation without validating their existence or types. A malformed request will produce an unhandled error (500) rather than a clean 400 response.",
      "suggestion": "Wrap `req.json()` in try/catch, validate that `body.packSlug`, `body.packName`, `body.action`, and `body.timestamp` are present and are strings before passing to the mutation.",
      "confidence": 0.85
    },
    {
      "id": "LOW-003",
      "title": "navigator.clipboard.writeText may fail without try/catch",
      "severity": "LOW",
      "category": "correctness",
      "file": "apps/explorer/components/dashboard/create-key-dialog.tsx:42-46",
      "description": "The `handleCopy` function calls `navigator.clipboard.writeText` with `await` but without a try/catch. This API can throw if the page doesn't have focus or clipboard permissions are denied, which would result in an unhandled promise rejection.",
      "suggestion": "Wrap in try/catch and show a fallback message if clipboard write fails.",
      "confidence": 0.8
    },
    {
      "id": "LOW-004",
      "title": "Silent error swallowing in constructs page and dashboard overview",
      "severity": "LOW",
      "category": "observability",
      "file": "apps/explorer/app/(dashboard)/dashboard/constructs/page.tsx:29",
      "description": "The `.catch(() => {})` pattern is used in multiple dashboard pages (constructs page line 29, dashboard overview lines 17, 25) to silently swallow all errors. Users see no feedback when API calls fail — the constructs page shows an empty table, and the overview shows stale or missing data.",
      "suggestion": "At minimum, set an error state and display a user-facing error message. Consider adding error logging for operational visibility.",
      "confidence": 0.75
    },
    {
      "id": "PRAISE-001",
      "title": "Clean Convex graceful degradation pattern",
      "severity": "PRAISE",
      "category": "architecture",
      "file": "apps/explorer/components/providers/convex-provider.tsx:14-21",
      "description": "The ConvexProvider cleanly falls back to passthrough rendering when NEXT_PUBLIC_CONVEX_URL is not configured. This means the entire app works without Convex, making the real-time features truly optional and improving developer experience for local development.",
      "suggestion": null
    },
    {
      "id": "PRAISE-002",
      "title": "API key shown only once pattern",
      "severity": "PRAISE",
      "category": "security",
      "file": "apps/api/src/routes/keys.ts:47-58",
      "description": "The full API key is returned only in the creation response, and the key hash (not the key itself) is stored in the database. The list endpoint returns only the prefix. This is the correct security pattern for API key management.",
      "suggestion": null
    },
    {
      "id": "PRAISE-003",
      "title": "Fire-and-forget webhook is non-blocking",
      "severity": "PRAISE",
      "category": "reliability",
      "file": "apps/api/src/routes/packs.ts:1300-1313",
      "description": "The Convex webhook call in the download handler is correctly fire-and-forget with `.catch(() => {})`, ensuring the real-time feed enhancement never impacts the critical download path.",
      "suggestion": null
    },
    {
      "id": "SPEC-001",
      "title": "Potential race condition in dashboard layout auth check",
      "severity": "SPECULATION",
      "category": "correctness",
      "file": "apps/explorer/app/(dashboard)/layout.tsx:21-25",
      "description": "The layout calls `initialize()` in one useEffect and checks `isLoading`/`isAuthenticated` in another. If `initialize()` is async and the second effect runs before initialization completes, there could be a brief window where `isLoading` is `false` (initial value before `initialize` sets it to `true`) and `isAuthenticated` is `false`, causing an unintended redirect. The middleware should catch most cases, but the client-side logic has this edge.",
      "suggestion": "Ensure `isLoading` starts as `true` (which it does in the store default) and that `initialize` doesn't briefly reset it. Verify the timing by checking if the redirect fires before initialization completes on slow connections.",
      "confidence": 0.4
    }
  ]
}
```


## Callouts

PR #155, opened 2026-03-11
feat(network): cycle-041 Network Cohesion

zkSoju:
## Summary
- **Category Derivation Pipeline**: Shared constants → DB column → API response → Explorer graph. Deletes 16-entry SLUG_CATEGORY_MAP hack.
- **Author DX**: `/skill-add` for two-phase construct growth, git-sync publish replacing stub, full publish SKILL.md with filesystem discovery
- **Network Automation**: `--register` flag in discovery script, pnpm→bun CI workflows, domain backfill script for 15 repos

## Sprints
| Sprint | Global ID | Tasks | Status |
|--------|-----------|-------|--------|
| Category Pipeline | sprint-45 | 10 | 9/10 (migration pending) |
| Author DX | sprint-46 | 4 | 4/4 |
| Network Automation | sprint-47 | 4 | 4/4 (backfill pending --apply) |

## Test plan
- [ ] `bun run --filter @loa-constructs/shared build` succeeds
- [ ] `bun tsc --noEmit --project apps/explorer/tsconfig.json` — 0 new errors
- [ ] Run migration on production: `bun -e` with 0011_packs_category.sql
- [ ] Re-seed: `bun seed:forge` populates category from domain[0]
- [ ] `GET /v1/constructs` returns real categories
- [ ] Explorer graph shows constructs distributed across 8 categories
- [ ] `./scripts/backfill-domain.sh` dry run reports 15 constructs
- [ ] `/skill-add test` creates 3 files in a construct directory

### DIFF
--- a/.claude/commands/skill-add.md
+++ b/.claude/commands/skill-add.md
@@ -1,0 +1,12 @@
+---
+agent: skill
+agent_path: .claude/skills/adding-skills/SKILL.md
+context_files:
+  - construct.yaml
+---
+
+# /skill-add
+
+Add a new skill to the current construct. Two-phase scaffolding: shell first (create), capabilities later (skill-add).
+
+Usage: `/skill-add <name>`
--- a/.claude/scripts/constructs-publish.sh
+++ b/.claude/scripts/constructs-publish.sh
@@ -312,9 +312,41 @@ do_push() {
         return $EXIT_AUTH_ERROR
     fi
 
-    # Upload (simplified — actual upload involves base64 encoding files)
-    print_status "Uploading $slug@$version to registry..."
-    print_warning "Publish upload not yet implemented — use registry web UI"
+    # Git-sync publish: push to origin + trigger registry sync (cycle-041)
+    print_status "Publishing $slug@$version via git-sync..."
+
+    # Push to origin with tags
+    if ! git push origin HEAD --tags 2>/dev/null; then
+        print_warning "Git push failed — check remote configuration"
+        print_warning "You may need to push manually: git push origin HEAD --tags"
+    fi
+
+    # Trigger sync via API
+    print_status "Triggering registry sync..."
+    local sync_tmp
+    sync_tmp=$(mktemp)
+    chmod 600 "$sync_tmp"
+
+    local sync_code
+    sync_code=$(curl --silent --proto '=https' --tlsv1.2 \
+        -o "$sync_tmp" -w "%{http_code}" \
+        -X POST \
+        -H "Authorization: Bearer $api_key" \
+        "${registry_url}/packs/${slug}/sync")
+
+    rm -f "$sync_tmp"
+
+    case "$sync_code" in
+        200|202)
+            print_success "Published $slug@$version — sync triggered successfully"
+            ;;
+        404)
+            print_warning "Pack '$slug' not found in registry — run 'bun seed:forge' to register"
+            ;;
+        *)
+            print_warning "Sync returned HTTP $sync_code — run 'bun seed:forge' to sync manually"
+            ;;
+    esac
 
     return $EXIT_SUCCESS
 }
--- a/.claude/skills/adding-skills/SKILL.md
+++ b/.claude/skills/adding-skills/SKILL.md
@@ -1,0 +1,108 @@
+# /skill-add — Add a Skill to a Construct
+
+## Purpose
+
+Two-phase scaffolding: `construct create` makes the shell, `/skill-add <name>` grows capabilities. Creates a dispatch-ready skill with context-aware content by reading existing skills.
+
+## Invocation
+
+```bash
+/skill-add <name>
+/skill-add research
+/skill-add deep-analysis
+```
+
+## Arguments
+
+| Argument | Description | Required |
+|----------|-------------|----------|
+| `name` | Skill name (lowercase, hyphens allowed) | Yes |
+
+## Workflow
+
+### Phase 1: Detect Construct Root
+
+1. Find `construct.yaml` in current directory or parent directories
+2. If not found, error: "Not in a construct directory. Run this from a directory containing construct.yaml."
+3. Read `construct.yaml` for construct name, type, description
+
+### Phase 2: Read Existing Skills for Context
+
+1. List `skills/*/index.yaml` to understand existing skill patterns
+2. Read 1-2 existing `SKILL.md` files to match style and voice
+3. Note trigger patterns, naming conventions, output paths
+
+### Phase 3: Create Skill Files
+
+Create three files:
+
+#### `skills/<name>/index.yaml`
+
+```yaml
+name: <name>
+version: "0.1.0"
+description: "<context-aware description based on construct domain>"
+
+triggers:
+  - pattern: "/<name>"
+    description: "<what this skill does>"
+
+entry: skills/<name>/SKILL.md
+```
+
+**Populate description and trigger description** based on:
+- The construct's domain (from construct.yaml or existing skills)
+- The skill name's semantic meaning
+- Existing skill patterns in this construct
+
+#### `skills/<name>/SKILL.md`
+
+Write a workflow stub that matches the style of existing skills in this construct. Include:
+- Purpose section
+- Invocation section with the trigger pattern
+- Workflow section with numbered steps (TODO placeholders)
+- Outputs table pointing to `grimoires/<construct-slug>/`
+
+#### `commands/<name>.md`
+
+```markdown
+---
+agent: skill
+agent_path: skills/<name>/SKILL.md
+context_files:
+  - construct.yaml
+---
+
+# /<name>
+
+Run the <name> skill.
+```
+
+### Phase 4: Report
+
+Print:
+- Files created
+- Next steps: "Edit `skills/<name>/SKILL.md` to define your workflow"
+
+## Guard Rails
+
+- **Error if not in construct directory**: Check for `construct.yaml`
+- **Error if skill exists**: Check `skills/<name>/` directory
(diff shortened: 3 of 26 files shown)


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/loa-constructs#155. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "SQL-INJECTION-1",
      "title": "SQL injection via string interpolation in raw SQL WHERE clause",
      "severity": "HIGH",
      "category": "security",
      "file": "apps/api/src/services/category.ts:174",
      "description": "The query `.where(sql`${packs.category} = ${normalizedSlug} AND ${packs.status} = 'published'`)` combines two conditions in a single `sql` template literal. While `${normalizedSlug}` is parameterized by Drizzle's `sql` tag, the `AND ${packs.status} = 'published'` part uses `${packs.status}` as a column reference which is fine, but the string `'published'` is embedded as a raw SQL literal. More critically, this mixes column references and value parameters in a single `sql` call in a way that differs from the skills query above (line 168) which uses `eq()`. The inconsistency is fragile — if `normalizedSlug` ever bypasses the normalizer (e.g., a code path change), it could become exploitable. The skills query on line 168 switched from `eq()` to `sql` template to avoid Drizzle enum constraints, but the packs query should use Drizzle's `and()` + `eq()` or at minimum two separate parameterized conditions.",
      "suggestion": "Use Drizzle's composable operators: `.where(and(sql`${packs.category} = ${normalizedSlug}`, eq(packs.status, 'published')))` to keep the raw SQL minimal and properly parameterized.",
      "confidence": 0.6
    },
    {
      "id": "MIGRATION-BACKFILL-1",
      "title": "Migration defaults unknown categories to 'development' silently — data integrity risk",
      "severity": "MEDIUM",
      "category": "data-integrity",
      "file": "apps/api/src/db/migrations/0011_packs_category.sql:15",
      "description": "The backfill UPDATE sets category to 'development' for any pack where `search_use_cases[1]` is NULL or doesn't match the known categories. This silently assigns a potentially incorrect category to packs that genuinely belong to other domains. If a pack has no `search_use_cases` data, 'development' may be wrong. There's no logging or tracking of how many rows got the default value, making it hard to audit after the fact.",
      "suggestion": "Consider leaving category as NULL for packs where no domain data exists, then using the seed script to properly derive categories. Alternatively, add a comment or separate query to count how many rows received the default so operators can verify.",
      "confidence": 0.7
    },
    {
      "id": "SEED-SQL-INJECTION-1",
      "title": "Seed script uses unparameterized string interpolation in SQL",
      "severity": "HIGH",
      "category": "security",
      "file": "scripts/seed-forge-packs.ts:613",
      "description": "The `category` value derived from `normalizeCategory()` is interpolated into the SQL template literal via `${category}`. Looking at the seed script pattern, it uses Drizzle's `sql` tagged template, so the value should be parameterized. However, the `normalizeCategory` function always returns one of the 8 canonical slugs or 'development' as default, so the blast radius is limited. Still, this is worth confirming — if the SQL is constructed as a raw string rather than via `sql` template tag, this would be a direct injection vector.",
      "suggestion": "Verify that the INSERT statement uses Drizzle's `sql` tagged template literal (not a plain string template) to ensure `${category}` is properly parameterized.",
      "confidence": 0.4
    },
    {
      "id": "NORMALIZE-DEFAULT-1",
      "title": "normalizeCategory silently defaults to 'development' for unknown slugs",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "packages/shared/src/categories.ts:59",
      "description": "The `normalizeCategory` function returns `'development'` for any unrecognized input. This is a silent lossy transformation. Consumers calling `normalizeCategory('some-typo')` will get `'development'` with no indication of an error. The previous implementation in `category.ts` returned the input unchanged for unknown slugs, so this is a behavioral change that could mask bugs (e.g., a construct.yaml with `domain: [devlopment]` typo would silently become 'development' — correct by accident, but `domain: [marketng]` would silently become 'development').",
      "suggestion": "Consider returning the input unchanged for non-legacy, unrecognized slugs and only defaulting at the call site, or at minimum log a warning when the default is applied. Alternatively, provide a separate `normalizeCategoryOrDefault()` function.",
      "confidence": 0.75
    },
    {
      "id": "BACKFILL-MACOS-SED-1",
      "title": "sed command in backfill-domain.sh uses GNU-specific syntax",
      "severity": "LOW",
      "category": "portability",
      "file": "scripts/backfill-domain.sh:79",
      "description": "The `sed` append syntax `sed \"/^description:/a\\\\\\ndomain:\\\\\\n  - ${domain}\"` uses a backslash-newline continuation that behaves differently on macOS (BSD sed) vs GNU sed. On macOS, `sed -i` requires a backup extension argument and the `\\n` in the append may not be interpreted correctly.",
      "suggestion": "Either add a note that GNU sed is required (CI uses Linux so this is fine for the workflow), or use `printf` + a temp file approach for cross-platform compatibility if developers are expected to run this locally on macOS.",
      "confidence": 0.8
    },
    {
      "id": "CACHE-STALE-1",
      "title": "Pack count changes won't invalidate category cache",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "apps/api/src/services/category.ts:72",
      "description": "The `listCategories` function now counts both skills and packs, but the cache key (`CACHE_KEYS.CATEGORIES_ALL`) and TTL remain unchanged. When packs are published or unpublished, the cached category counts will be stale until TTL expiry. Previously this only affected skill counts, but now pack mutations (which may be more frequent during seeding) will also cause stale data.",
      "suggestion": "Ensure that pack publish/unpublish operations invalidate the categories cache, or document the accepted staleness window.",
      "confidence": 0.6
    },
    {
      "id": "WORKFLOW-AUTO-DISCOVER-DEFAULT",
      "title": "Auto-discover default changed from 'true' to 'false' — may break scheduled sync",
      "severity": "MEDIUM",
      "category": "operations",
      "file": ".github/workflows/sync-constructs.yml:63",
      "description": "The summary line changed the default for auto-discover from `'true'` to `'false'`. However, looking at line 53-55, the `--auto-discover` flag is only added when the input is explicitly `'true'`. On scheduled runs (`schedule` trigger), `github.event.inputs.auto_discover` is undefined, so the flag won't be passed. If the previous behavior was to auto-discover on schedule, this is a regression. The summary line change suggests this was intentional, but it's worth verifying.",
      "suggestion": "Confirm whether scheduled runs should auto-discover. If yes, the default in the conditional on line 53 should also default to 'true' for schedule events.",
      "confidence": 0.65
    },
    {
      "id": "SHARED-EXPORT-1",
      "title": "Shared package exports categories without updating package.json exports map",
      "severity": "MEDIUM",
      "category": "build",
      "file": "packages/shared/src/index.ts:8",
      "description": "The new `categories.ts` module is re-exported from the shared package's `index.ts`, but the diff doesn't show any changes to `packages/shared/package.json` or `tsconfig.json`. If the shared package uses explicit `exports` field in package.json (common in monorepo setups), the new module may not be resolvable by consumers unless the barrel export through `index.ts` is already covered.",
      "suggestion": "Verify that `packages/shared/package.json` exports configuration covers the new `categories.js` module, either through the barrel export or explicit path mapping.",
      "confidence": 0.5
    },
    {
      "id": "PRAISE-SINGLE-SOURCE",
      "title": "Excellent consolidation of category definitions into shared package",
      "severity": "PRAISE",
      "category": "architecture",
      "file": "packages/shared/src/categories.ts:1",
      "description": "Moving the category taxonomy, legacy mappings, and normalizeCategory function into a shared package eliminates 3+ copies of the same data across the API, explorer, and seed scripts. The type-safe `CategorySlug` union type and `isValidCategory` guard are well-designed. This is a meaningful DRY improvement.",
      "suggestion": null
    },
    {
      "id": "PRAISE-MIGRATION-DESIGN",
      "title": "Good choice using VARCHAR over ENUM for category column",
      "severity": "PRAISE",
      "category": "database",
      "file": "apps/api/src/db/migrations/0011_packs_category.sql:9",
      "description": "Using `VARCHAR(50)` instead of a PostgreSQL ENUM type for the category column is a pragmatic choice explicitly called out in the migration comment. Adding values to PostgreSQL ENUMs requires `ALTER TYPE` which can be disruptive and has transactional limitations. This allows the category taxonomy to evolve without migration pain.",
      "suggestion": null
    },
    {
      "id": "BACKFILL-RACE-1",
      "title": "Backfill script may create duplicate PRs or fail on existing branch",
      "severity": "LOW",
      "category": "operations",
      "file": "scripts/backfill-domain.sh:88",
      "description": "The branch name `chore/domain-backfill` is hardcoded and the `gh api` call to create the ref uses `|| true` to suppress errors. If the script is run twice with `--apply`, the second run will silently fail to create the branch (it already exists) but will still try to update the file and create a PR, potentially causing confusing error states.",
      "suggestion": "Check if the branch already exists before creating it, or delete+recreate, or use a timestamped branch name like `chore/domain-backfill-$(date +%s)`.",
      "confidence": 0.8
    },
    {
      "id": "EXPLORER-MISSING-SORT-1",
      "title": "Default categories in explorer lose sortOrder property",
      "severity": "LOW",
      "category": "correctness",
      "file": "apps/explorer/lib/data/fetch-categories.ts:18",
      "description": "The `DEFAULT_CATEGORIES` mapping from `CATEGORIES` drops the `sortOrder` property (not included in the mapped object). The `Category` type from `@/lib/types/graph` may or may not include `sortOrder`. If any consumer relies on sort order from the default categories, the ordering would be lost.",
      "suggestion": "Include `sortOrder: cat.sortOrder` in the mapped object if the Category type supports it, or verify no consumer needs it from the defaults.",
      "confidence": 0.5
    }
  ]
}
```


## Callouts

PR #156, opened 2026-03-12
feat(api): cycle-042 API test stability

zkSoju:
## Summary

- Delete ~1,790 lines of fake tests that assert on locally-defined constants (zero HTTP calls, zero real code exercised)
- Replace with **334 real tests** across 22 test files covering all 11 fragile areas that cause API downtime
- Create shared test helpers (`mock-db`, `mock-redis`, `fixtures`, `auth`) reducing per-test boilerplate to <10 lines
- Fix explorer Vercel build failure (`@loa-constructs/shared` workspace dependency)

### Stability Tests (Tier 1) — 7 fragile areas covered
- **Blacklist fail-secure**: Redis outage returns `true` (blocks token use) — prevents silent auth bypass
- **Rate limiter fail-open/closed**: Auth endpoints 503 on Redis error, non-auth pass through
- **Auth middleware**: 12 cases — requireAuth, optionalAuth, requireTier, requireOrgMember, requireVerifiedEmail
- **Error handler**: AppError structured response vs unknown error (no stack leak), duck-typing check
- **Health readiness**: DB up/down, Redis down/not-configured probe behavior
- **Auth routes**: Login success/fail, refresh, blacklist, logout, me, validate

### Core API Tests (Tier 2)
- Constructs: list, filters, pagination, Zod validation, detail, 404, HEAD, summary, visibility
- Webhooks: Stripe/GitHub HMAC verification, idempotency, replay protection
- Categories: response shape contract, sort order

### Contract Tests (Tier 3)
- Snapshot tests expanded to all public endpoints
- Zod response schema validation (health, error, paginated, category)

### Config Changes
- `testTimeout: 10000` (bcrypt operations)
- `pool: 'forks'` (isolate test files, prevent mock pollution)
- Coverage thresholds: `statements: 40, branches: 30, functions: 35, lines: 40`
- UPSTASH env guards in vitest.setup.ts

## Test plan
- [x] `cd apps/api && bun run test` — 334 tests passing
- [x] 8 pre-existing failures in `category.test.ts` (imports non-existent `DEFAULT_CATEGORIES` — not from this PR)
- [ ] Verify CI passes
- [ ] Intentionally break auth middleware → stability tests catch it
- [ ] Intentionally break constructs response shape → contract tests catch it


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/loa-constructs#156. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "FIND-001",
      "title": "Shared select chain instance causes test cross-contamination in auth.test.ts middleware",
      "severity": "MEDIUM",
      "category": "Testing",
      "file": "apps/api/src/middleware/auth.test.ts:14-27",
      "description": "The `selectChain` and `updateChain` objects at lines 14-27 are defined once as module-level singletons, but the `then` callback captures `selectResults` and `selectCallIndex` by closure reference. Because the same chain object is reused across all `db.select()` calls, if two tests run without proper reset, the shared `selectCallIndex` can drift. The auth.test.ts file creates fresh chains inside `db.select` (line 33-43) which is correct, but the outer `selectChain` at line 14 is never used — it's dead code that could confuse future maintainers.",
      "suggestion": "Remove the unused outer `selectChain` (lines 14-27) and `updateChain` (lines 10-14) since the inner factory in `db.select` already creates fresh chains per call.",
      "confidence": 0.7
    },
    {
      "id": "FIND-002",
      "title": "Index-based mock result consumption is fragile and order-dependent",
      "severity": "MEDIUM",
      "category": "Maintainability",
      "file": "apps/api/src/routes/auth.test.ts:7-13",
      "description": "All route test files (auth.test.ts, constructs.test.ts, webhooks.test.ts) use a global `selectCallIndex` that increments on every `db.select()` resolution. This means tests must know the exact number and order of internal DB calls made by the production code. Any refactor that adds/removes/reorders a DB query silently breaks tests — they'll consume the wrong mock result without a clear error message, producing confusing failures.",
      "suggestion": "Consider a mock DB approach where results are keyed by table name or query pattern rather than call order. The `mock-db.ts` helper already uses terminal mocks (`_selectResult`) which is slightly better, but even that shares state. A per-table result queue would be more resilient to implementation changes.",
      "confidence": 0.8
    },
    {
      "id": "FIND-003",
      "title": "Blacklist service fail-secure behavior returns true on Redis error — good security design",
      "severity": "PRAISE",
      "category": "Security",
      "file": "apps/api/src/services/blacklist.test.ts:87-92",
      "description": "The `isBlacklisted()` test confirms that when Redis errors, the function returns `true` (fail-secure). This means if the blacklist store is unavailable, tokens are treated as revoked rather than valid — the correct security posture for token revocation.",
      "suggestion": null,
      "confidence": 0.95
    },
    {
      "id": "FIND-004",
      "title": "Rate limiter auth endpoint fail-closed behavior is properly tested",
      "severity": "PRAISE",
      "category": "Security",
      "file": "apps/api/src/middleware/rate-limiter.test.ts:115-123",
      "description": "The test validates that auth endpoints return 503 when Redis is unavailable (fail-closed), while non-auth endpoints fail open with a degradation header. This dual-mode behavior correctly protects authentication paths from brute-force during infrastructure failures.",
      "suggestion": null,
      "confidence": 0.95
    },
    {
      "id": "FIND-005",
      "title": "Removed tests were testing hardcoded data, not actual API behavior",
      "severity": "PRAISE",
      "category": "Testing",
      "file": "apps/api/tests/e2e/constructs.test.ts",
      "description": "The deleted `constructs.test.ts` (713 lines), `pack-flow.test.ts` (453 lines), `creator.test.ts` (260 lines), `constructs.test.ts` service (370 lines), `submissions.test.ts` (290 lines), and `skills.test.ts` (147 lines) were largely testing hardcoded response objects and local variable assertions — not actually exercising routes. The replacements use `app.request()` to make real HTTP calls through the Hono app, which is a significant quality improvement.",
      "suggestion": null,
      "confidence": 0.95
    },
    {
      "id": "FIND-006",
      "title": "JWT secret hardcoded in test helpers must match vitest.setup.ts",
      "severity": "LOW",
      "category": "Testing",
      "file": "apps/api/tests/helpers/auth.ts:14-15",
      "description": "The `TEST_JWT_SECRET` and `TEST_JWT_ISSUER` are hardcoded strings that must match what's in `vitest.setup.ts` and every test file's `vi.mock('../config/env.js')`. If any of these diverge, JWT verification will silently fail. There's no single source of truth — the secret appears in at least 6 different files.",
      "suggestion": "Extract the test JWT secret and issuer into a single shared constant (e.g., in `tests/helpers/constants.ts`) and import it everywhere, including in `vi.mock` factories.",
      "confidence": 0.85
    },
    {
      "id": "FIND-007",
      "title": "process.env mutation in webhooks test may leak to other test files",
      "severity": "MEDIUM",
      "category": "Testing",
      "file": "apps/api/src/routes/webhooks.test.ts:229",
      "description": "`process.env.GITHUB_WEBHOOK_SECRET = GITHUB_SECRET;` is set at module scope with no cleanup. With `pool: 'forks'` in vitest config, each file gets its own process so this is currently safe, but if the pool config changes to `threads`, this mutation would leak across test files.",
      "suggestion": "Move the `process.env` assignment into `beforeEach` and restore the original value in `afterEach`, or add a comment documenting the dependency on `pool: 'forks'`.",
      "confidence": 0.6
    },
    {
      "id": "FIND-008",
      "title": "Coverage thresholds are set very low for a test stability PR",
      "severity": "LOW",
      "category": "Testing",
      "file": "apps/api/vitest.config.ts:16-21",
      "description": "Coverage thresholds are set to 40/30/35/40 (statements/branches/functions/lines). For a PR focused on test stability, these thresholds are quite low and may not catch regressions effectively. However, this may be intentional as a starting point.",
      "suggestion": "Consider raising thresholds incrementally as test coverage improves, or document the target thresholds in a tracking issue.",
      "confidence": 0.5
    },
    {
      "id": "FIND-009",
      "title": "Shared test helpers (mock-db.ts, mock-redis.ts) are created but not used by any test in this PR",
      "severity": "LOW",
      "category": "Maintainability",
      "file": "apps/api/tests/helpers/mock-db.ts:1",
      "description": "The `createMockDb`, `getMockDbModule`, `createMockRedis`, `getMockRedisModule` helpers and the `fixtures.ts` factories are exported from `tests/helpers/index.ts`, but none of the test files in this PR actually import from these helpers (except `auth.ts`). Each test file instead creates its own inline mock chains, leading to significant duplication across auth.test.ts, constructs.test.ts, health.test.ts, webhooks.test.ts, etc.",
      "suggestion": "Refactor at least one test file to use the shared helpers to validate the pattern works, then migrate the others incrementally. Currently the helpers are untested dead code.",
      "confidence": 0.9
    },
    {
      "id": "FIND-010",
      "title": "Massive mock boilerplate duplicated across every test file (~60-80 lines each)",
      "severity": "MEDIUM",
      "category": "Maintainability",
      "file": "apps/api/src/routes/constructs.test.ts:15-130",
      "description": "The `vi.mock('../db/index.js')`, `vi.mock('../config/env.js')`, `vi.mock('../lib/logger.js')`, `vi.mock('../services/redis.js')`, etc. blocks are nearly identical across auth.test.ts, constructs.test.ts, health.test.ts, webhooks.test.ts, categories.test.ts, and the contract tests. This is ~80 lines of boilerplate per file. The shared helpers in `tests/helpers/` were seemingly created to solve this but aren't used.",
      "suggestion": "Create a shared setup module (e.g., `tests/helpers/setup-mocks.ts`) that configures all common mocks, and call it from each test file. Vitest's `setupFiles` could also help for truly global mocks.",
      "confidence": 0.85
    },
    {
      "id": "FIND-011",
      "title": "Response schema test validates Zod error shape loosely",
      "severity": "LOW",
      "category": "Testing",
      "file": "apps/api/tests/contract/response-schemas.test.ts:343-349",
      "description": "The Zod validation error test at line 343-349 only asserts `expect(body).toHaveProperty('error')` without validating against the `ErrorResponseSchema`. This is weaker than other tests in the same file that use full schema validation. If the Zod error format changes, this test won't catch the regression.",
      "suggestion": "Validate the Zod error response against the `ErrorResponseSchema` or a dedicated `ValidationErrorResponseSchema` for consistency.",
      "confidence": 0.75
    },
    {
      "id": "FIND-012",
      "title": "Anti-enumeration test for login is well-designed",
      "severity": "PRAISE",
      "category": "Security",
      "file": "apps/api/src/routes/auth.test.ts:194-207",
      "description": "The test validates that a non-existent email returns the same error shape and message as a wrong password ('Invalid email or password'), confirming anti-enumeration behavior. This is a good security practice to prevent attackers from discovering valid email addresses.",
      "suggestion": null,
      "confidence": 0.95
    },
    {
      "id": "FIND-013",
      "title": "Webhook HMAC test uses real crypto — good contract coverage",
      "severity": "PRAISE",
      "category": "Testing",
      "file": "apps/api/src/routes/webhooks.test.ts:231-237",
      "description": "The GitHub webhook tests compute real HMAC-SHA256 signatures using Node's `createHmac`, then send them through the full Hono app request pipeline. This validates the actual signature verification code path rather than mocking it away, providing strong contract coverage.",
      "suggestion": null,
      "confidence": 0.95
    },
    {
      "id": "FIND-014",
      "title": "testTimeout increased to 10s without justification",
      "severity": "LOW",
      "category": "Testing",
      "file": "apps/api/vitest.config.ts:9",
      "description": "The `testTimeout: 10000` is 2x the default (5000ms). While this may be needed for bcrypt hashing in auth tests (which uses cost factor 4), it could mask slow tests. The bcrypt cost factor of 4 in tests should be fast enough for the default timeout.",
      "suggestion": "Add a comment explaining why 10s is needed, or reduce back to default and use per-test timeouts for known slow tests.",
      "confidence": 0.5
    },
    {
      "id": "FIND-015",
      "title": "Explorer package.json adds shared workspace dependency",
      "severity": "LOW",
      "category": "Dependencies",
      "file": "apps/explorer/package.json:49",
      "description": "Adding `@loa-constructs/shared: workspace:*` to the explorer is noted here. The `transpilePackages` addition in `next.config.ts` correctly handles the CJS/ESM bridging. This appears intentional but is unrelated to the PR title about API test stability.",
      "suggestion": "Consider separating the explorer dependency change into its own PR for clearer change tracking, unless it was needed to fix a build issue caused by other changes in this PR.",
      "confidence": 0.5
    }
  ]
}
```


## Callouts

PR #165, opened 2026-03-13
feat(cycle-048): Construct Discovery — API enrichment, browse-first explorer

zkSoju:
## Summary

Full cycle-048 implementation: Construct Discovery — Browse-First Navigation, Search, Composability.

### Sprint 1: API Enrichment (2 commits)
- `domains[]`, `expertiseSummary[]`, `skillDetails[]`, `composeWith[]`, `dependedBy[]` in API response
- `domain` query param for filtering
- Relevance scoring: domains (0.8), expertise (0.9), skill slugs (0.7)
- Seed script extracts skill descriptions from SKILL.md

### Sprint 3: Explorer Browse-First UX (1 commit)
- Flat catalog — all 23 constructs visible, no pagination
- Fuse.js client-side instant search across enriched fields
- Domain tag pills with stackable AND filtering
- 5 intent entry cards (Create, Validate, Launch, Research, Direct)
- Card enhancement: skill count, domain chips, composition degree
- Works-with panel on construct detail pages
- URL state sync (?q=, ?tags=, ?intent=)

### Sprint 2: `/constructs` CLI (separate PR)
Shipped as PR to upstream Loa repo (0xHoneyJar/loa).

### Context
- Design doc: `grimoires/gecko/construct-discovery-design.md`
- 3 research agents, 155 Gemini grounded queries
- 23-construct surface analysis (composability graph, vocabulary, gaps)
- References: skills.sh (Vercel), Google Maps AI, Vercel DX, npms.io, QUEST framework

## Test plan

- [ ] `GET /v1/constructs?per_page=100` returns enriched fields for all constructs
- [ ] `GET /v1/constructs?domain=security` filters correctly
- [ ] `GET /v1/constructs?q=incident` ranks Hardening higher (expertise match)
- [ ] Explorer `/constructs` shows flat grid with intent cards and domain pills
- [ ] Fuse.js search filters in <100ms
- [ ] Detail page shows "Works with" panel
- [ ] URL params (?q=, ?tags=) produce correct filtered view
- [ ] Existing clients unaffected (new fields additive)
- [ ] Run seed — skill descriptions populate


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/loa-constructs#165. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "FIND-001",
      "title": "Domain filter applied after pagination, causing incorrect result counts and short pages",
      "severity": "HIGH",
      "category": "correctness",
      "file": "apps/api/src/routes/constructs.ts",
      "description": "The `domain` query filter is applied *after* `listConstructs` has already paginated results (line 191-195). This means: (1) the returned `pagination.total` count is wrong since it doesn't account for domain filtering, (2) pages may have fewer items than `per_page`, and (3) some matching constructs on later DB pages will never be returned. The filter needs to be pushed down into the query layer or applied before pagination.",
      "suggestion": "Pass the `domain` filter into `listConstructs` so it's applied at the database/query level before pagination, or at minimum recalculate `total` after filtering and adjust pagination metadata accordingly.",
      "confidence": 0.95
    },
    {
      "id": "FIND-002",
      "title": "O(n²) dependedBy computation on every list request",
      "severity": "MEDIUM",
      "category": "performance",
      "file": "apps/api/src/services/constructs.ts",
      "description": "At line 674, `computeDependedByMap` iterates over all returned constructs and their dependencies on every `listConstructs` call. The cross-reference is inherently limited to the current page of results — a construct's `dependedBy` will only include other constructs in the same response page, not the full corpus. This makes the field both misleading (incomplete data) and wasteful (computed per request).",
      "suggestion": "Pre-compute the dependedBy map at seed/publish time and store it, or compute it across the full dataset and cache it with a TTL, rather than computing a partial view per paginated request.",
      "confidence": 0.9
    },
    {
      "id": "FIND-003",
      "title": "Relevance score accumulation can exceed meaningful bounds with enrichment fields",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "apps/api/src/services/constructs.ts",
      "description": "Lines 227-249 add up to +0.8, +0.9, and +0.7 per matching term for domains, expertise, and skills respectively. With multiple query terms, these scores accumulate additively without any cap or normalization. A multi-term query like 'react typescript frontend' could inflate the score by 7.2+ from enrichment fields alone (3 terms × 2.4 per term), potentially dwarfing other signals and making ranking unpredictable.",
      "suggestion": "Add each enrichment category's bonus at most once regardless of how many terms match, or normalize/cap the final score. Consider using `break` after the first matching term per category.",
      "confidence": 0.8
    },
    {
      "id": "FIND-004",
      "title": "Bidirectional substring matching in relevance scoring is overly broad",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "apps/api/src/services/constructs.ts",
      "description": "Lines 228, 237, 246 use `d.includes(term) || term.includes(d)`. The `term.includes(d)` direction means a short domain/expertise/skill value like 'ai' would match any query term containing 'ai' (e.g., 'maintain', 'contains', 'explain'). This will produce many false-positive relevance boosts.",
      "suggestion": "Remove the `term.includes(d)` direction, or require exact word boundary matching. At minimum, only allow the forward direction `d.includes(term)` to avoid short-string false positives.",
      "confidence": 0.85
    },
    {
      "id": "FIND-005",
      "title": "extractExpertiseSummary type assertion on unvalidated external data",
      "severity": "MEDIUM",
      "category": "correctness",
      "file": "apps/api/src/services/constructs.ts",
      "description": "Line 502: `const domains = identity.expertiseDomains as Array<{ name: string; depth?: number }>` performs an unchecked type assertion on data that comes from a database row. If the shape doesn't match (e.g., it's a plain string array or has different field names), the `.filter(d => d && d.name ...)` may silently produce empty results or throw on unexpected shapes. The `try/catch` mitigates crashes but masks data issues.",
      "suggestion": "Add runtime validation (e.g., check `typeof d.name === 'string'` inside the filter) to handle unexpected shapes gracefully and log warnings for data quality.",
      "confidence": 0.7
    },
    {
      "id": "FIND-006",
      "title": "Moodboard embed.ts writes file after every single embedding",
      "severity": "LOW",
      "category": "performance",
      "file": "apps/explorer/scripts/moodboard/embed.ts",
      "description": "Line 145: `saveEmbeddings(IMAGES_OUTPUT, entries)` is called inside the for-loop after each successful embed, and then again at line 152 after the loop. For large batches this results in O(n²) total bytes written (serializing the growing array n times). The intent (crash recovery) is good but the implementation is expensive.",
      "suggestion": "Write the partial checkpoint less frequently (e.g., every 10 entries) or write to a temp file and rename at the end, keeping the resume logic but reducing I/O.",
      "confidence": 0.75
    },
    {
      "id": "FIND-007",
      "title": "GEMINI_API_KEY checked independently in both embed functions instead of once at startup",
      "severity": "LOW",
      "category": "code-quality",
      "file": "apps/explorer/scripts/moodboard/embed.ts",
      "description": "Both `embedImages` (line 108) and `embedTokens` (line 156) independently check for `GEMINI_API_KEY`. The key should be validated once before dispatching to either function.",
      "suggestion": "Validate `GEMINI_API_KEY` in the CLI parsing section before calling either function.",
      "confidence": 0.6
    },
    {
      "id": "FIND-008",
      "title": "New API response fields exposed without versioning consideration",
      "severity": "LOW",
      "category": "api-design",
      "file": "apps/api/src/routes/constructs.ts",
      "description": "Lines 88-93 add five new fields (`domains`, `expertise_summary`, `skill_details`, `compose_with`, `depended_by`) to the constructs API response. These are always returned (defaulting to empty arrays) with no feature flag or API version gating. Since `depended_by` is computed incorrectly (only within current page), consumers may build on unreliable data.",
      "suggestion": "Consider gating these behind an `?enrich=true` query parameter or an API version header until the data quality (especially `depended_by`) is validated.",
      "confidence": 0.65
    },
    {
      "id": "FIND-009",
      "title": "Accessibility: aria-hidden on visible text labels",
      "severity": "LOW",
      "category": "accessibility",
      "file": "apps/explorer/components/dashboard/sidebar.tsx",
      "description": "Section labels like 'sector::operations' (line 40) and 'sector::admin' (line 71) use `aria-hidden=\"true\"`, which hides them from screen readers. While these are decorative section headers, they provide the only visual indication of navigation grouping. Similarly in dashboard-header.tsx, the breadcrumb separator and badge labels are hidden. Screen reader users lose navigation context.",
      "suggestion": "Use `role=\"heading\"` with `aria-level` for section labels, or provide a visually-hidden equivalent. Keep `aria-hidden` only on purely decorative elements.",
      "confidence": 0.7
    },
    {
      "id": "FIND-010",
      "title": "Google Generative AI SDK added as devDependency for a script tool",
      "severity": "LOW",
      "category": "dependencies",
      "file": "apps/explorer/package.json",
      "description": "`@google/generative-ai` is added as a devDependency. This is correct for a script-only tool, but `glob` (also added as devDependency) was upgraded from v10 to v13 in the lockfile's top-level resolution, which may affect other consumers of `glob` in the monorepo.",
      "suggestion": "Verify the glob v10→v13 resolution change doesn't break other workspace packages. The lockfile shows test-exclude now gets its own glob@10.5.0 resolution, so it should be fine, but worth a CI check.",
      "confidence": 0.5
    },
    {
      "id": "FIND-011",
      "title": "Clean extraction of reusable StatCard and QuickLink components",
      "severity": "PRAISE",
      "category": "code-quality",
      "file": "apps/explorer/components/dashboard/stat-card.tsx",
      "description": "Extracting `StatCard` and `QuickLink` from the dashboard page into their own component files with proper TypeScript interfaces, loading states, variant support, and accessibility labels is well-done. The loading state detection via `value === '...'` is simple and effective.",
      "suggestion": null
    },
    {
      "id": "FIND-012",
      "title": "Moodboard embedding tool with resume support is thoughtfully designed",
      "severity": "PRAISE",
      "category": "code-quality",
      "file": "apps/explorer/scripts/moodboard/embed.ts",
      "description": "The partial resume logic that loads existing embeddings and skips already-processed entries, combined with exponential backoff retry, makes this a robust CLI tool for an operation that involves external API calls that can fail.",
      "suggestion": null
    },
    {
      "id": "FIND-013",
      "title": "SKILL.md description extraction uses only first non-header, non-frontmatter line",
      "severity": "LOW",
      "category": "correctness",
      "file": "scripts/seed-forge-packs.ts",
      "description": "Lines 517-530: The frontmatter parser toggles `inFrontmatter` on every `---` line. If a SKILL.md has no frontmatter, the first `---` (e.g., a horizontal rule) would set `inFrontmatter = true` and the second would toggle it back, causing content between them to be skipped. Also, the description is truncated to 120 chars without checking for mid-word truncation.",
      "suggestion": "Consider checking for frontmatter only at the start of the file (first line must be `---`), and add word-boundary-aware truncation.",
      "confidence": 0.6
    }
  ]
}
```


## Callouts

PR #169, opened 2026-03-14
security: pre-share audit fixes

zkSoju:
## Pre-Share Security Audit

2-agent audit (API surface + secrets/data exposure). Fixes the most critical findings before sharing the network.

### Fixed

| ID | Severity | Fix |
|---|---|---|
| CRIT-2 | Critical | Dashboard middleware validates JWT structure + expiry, not just cookie existence |
| CRIT-3 | Critical | `/audit/admin` checks `user.role` not `user.tier` — enterprise users no longer see all logs |
| HIGH-5 | High | `.run/` state files untracked (526 lines removed from git) |
| HIGH-6 | High | `/health/metrics` stripped of process PID, Node version, platform, arch |
| — | High | BFF logout sends refresh_token to API — tokens now actually blacklisted on logout |

### Deferred (follow-up session)

| ID | Severity | Issue |
|---|---|---|
| CRIT-1 | Critical | OAuth tokens in redirect query string (larger redesign — BFF code exchange) |
| CRIT-4 | Critical | Cron reconcile uses CRON_SECRET as JWT (always fails silently) |
| HIGH-1 | High | Rate limiter fails open when Redis unavailable |
| HIGH-4 | High | Single CONVEX_WRITE_KEY shared across all surfaces |

### What's solid (no changes needed)
CORS, SQL injection (parameterized), Stripe webhook verification, GitHub webhook HMAC, API key auth, admin middleware, Dynamic Labs JWKS.


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/loa-constructs#169. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "SEC-01",
      "title": "JWT payload decoded with atob() which does not handle base64url correctly for all inputs",
      "severity": "MEDIUM",
      "category": "Security",
      "file": "apps/explorer/middleware.ts",
      "description": "The code replaces `-` with `+` and `_` with `/` (correct base64url→base64 conversion), but `atob()` also requires proper padding (`=`). Many JWT libraries omit padding from base64url segments. If the payload segment length is not a multiple of 4, `atob()` will throw in some runtimes, causing all tokens with unpadded payloads to be treated as malformed and deleted. This could cause legitimate users to be logged out.",
      "suggestion": "Add padding before calling `atob()`: `const base64 = parts[1].replace(/-/g, '+').replace(/_/g, '/'); const padded = base64.padEnd(base64.length + (4 - base64.length % 4) % 4, '='); const payload = JSON.parse(atob(padded));`",
      "confidence": 0.75
    },
    {
      "id": "SEC-02",
      "title": "Logout now requires both accessToken AND refreshToken, breaking partial-state logout",
      "severity": "MEDIUM",
      "category": "Security",
      "file": "apps/explorer/app/api/auth/logout/route.ts",
      "description": "The condition changed from `if (accessToken)` to `if (accessToken && refreshToken)`. If the refresh_token cookie is missing (e.g., expired, cleared by the browser, or a different-domain issue), the server-side logout is silently skipped entirely. The refresh token remains valid on the server and could be replayed if obtained by an attacker. The old code at least attempted server-side token invalidation with just the access token.",
      "suggestion": "Consider making the condition `if (accessToken || refreshToken)` and conditionally including each in the request, or at minimum attempt server-side logout if either token is present. The API should handle missing refresh_token gracefully.",
      "confidence": 0.85
    },
    {
      "id": "SEC-03",
      "title": "Expired JWT in middleware allows passthrough without any cache-busting or header signal",
      "severity": "LOW",
      "category": "Security",
      "file": "apps/explorer/middleware.ts",
      "description": "When the JWT is expired (line 30-33), the middleware calls `NextResponse.next()` to let the client-side refresh flow handle it. However, there's no signal (e.g., a response header) to the client that the token is known-expired, nor is the expired cookie cleared. If the client-side refresh flow has a bug or race condition, the user will repeatedly hit APIs with an expired token until the cookie naturally expires.",
      "suggestion": "Consider setting a response header like `X-Token-Expired: true` on the NextResponse so client-side code can eagerly trigger refresh, or redirect to a dedicated refresh endpoint.",
      "confidence": 0.4
    },
    {
      "id": "SEC-04",
      "title": "Admin role check relies on user.role field without validating its source",
      "severity": "MEDIUM",
      "category": "Security",
      "file": "apps/api/src/routes/audit.ts",
      "description": "The authorization check `ADMIN_ROLES.includes(user.role || 'user')` depends on `user.role` being trustworthy. If `user.role` is derived from the JWT payload without server-side verification (e.g., from a decoded but unverified claim), this is bypassable. The fallback `|| 'user'` is correct defensive coding, but the overall security depends on where `user` is populated. Additionally, using `includes()` on a small array with a potentially undefined/null value is fine, but the type of `user.role` should be validated as a string.",
      "suggestion": "Verify that `user.role` is set from a trusted server-side source (database lookup or verified JWT claim) and not from unverified client input. Consider adding a type guard to ensure `user.role` is a string before the `includes()` call.",
      "confidence": 0.5
    },
    {
      "id": "PRAISE-01",
      "title": "Removal of process fingerprinting data from health endpoint",
      "severity": "PRAISE",
      "category": "Security",
      "file": "apps/api/src/routes/health.ts",
      "description": "Removing `process.pid`, `process.version`, `process.platform`, and `process.arch` from the metrics endpoint is a solid hardening measure. This data directly aids attackers in fingerprinting the runtime for targeted CVE exploitation.",
      "suggestion": null
    },
    {
      "id": "PRAISE-02",
      "title": "JWT structural validation in middleware is a good defense-in-depth measure",
      "severity": "PRAISE",
      "category": "Security",
      "file": "apps/explorer/middleware.ts",
      "description": "Adding structural JWT validation (3-segment check, payload parse) in the middleware before requests reach the API is good defense-in-depth. It catches obviously invalid tokens early and clears them, improving UX for corrupted cookie scenarios.",
      "suggestion": null
    },
    {
      "id": "SEC-05",
      "title": "Audit admin authorization changed from tier-based to role-based without migration consideration",
      "severity": "LOW",
      "category": "Correctness",
      "file": "apps/api/src/routes/audit.ts",
      "description": "The access control changed from `user.tier !== 'enterprise'` to `!ADMIN_ROLES.includes(user.role || 'user')`. If existing enterprise users were relying on the old behavior to access audit logs, this is a breaking change. Conversely, the old check was clearly wrong (enterprise ≠ admin), so the fix is correct — but any existing admin workflows that relied on enterprise tier will break.",
      "suggestion": "Ensure that actual administrators have the 'admin' or 'super_admin' role set in the database before deploying this change. Consider a data migration or at minimum a verification step.",
      "confidence": 0.6
    },
    {
      "id": "SEC-06",
      "title": "Refresh token sent in request body to logout endpoint — verify API expects this format",
      "severity": "LOW",
      "category": "Correctness",
      "file": "apps/explorer/app/api/auth/logout/route.ts",
      "description": "The logout request now sends `{ refresh_token: refreshToken }` in the body. This assumes the backend API at `/auth/logout` expects a JSON body with this exact field name. If the API expects a different field name (e.g., `refreshToken` camelCase) or expects the refresh token in a different location (cookie, header), this will silently fail since errors are caught and ignored.",
      "suggestion": "Verify that the API's logout endpoint schema matches the field name `refresh_token` and that it processes the JSON body correctly.",
      "confidence": 0.4
    }
  ]
}
```


## Callouts

PR #173, opened 2026-03-15
feat: analytics pipeline + GEO optimization (cycle-051)

zkSoju:
## Summary

- **Umami Cloud** analytics tracking on `constructs.network` (gated behind env var, cookieless, privacy-preserving)
- **Telegram digest bot** — daily anomaly-first analytics digest via Convex cron (14:00 UTC) to signal group with Lily
- **JSON-LD structured data** on all construct detail pages (`SoftwareApplication`) and root layout (`WebSite`)
- **About page enrichment** — Organization JSON-LD, live construct/skill counts, "Built by 0xHoneyJar" attribution

## Context

Cycle-051 PRD: `grimoires/loa/prd.md`. Research artifacts in `grimoires/loa/context/analytics-architecture/` (gitignored — 5 documents, 917 lines of research including adversarial review).

Key decisions:
- Umami Cloud over GA4 (no cookies, privacy policy stays accurate, full REST API for TG digest)
- Umami Cloud over self-hosted (avoids Prisma/PgBouncer conflict with Supabase)
- Anomaly-first digest design (lead with changes, one-liner if steady state)
- Single Telegram bot shared with social oracle construct

Connected to: beacon issue #3 (Lily's GEO research) — analytics feeds the traffic→GEO→optimization loop.

## Changes

| File | Change |
|------|--------|
| `app/layout.tsx` | Umami script + WebSite JSON-LD |
| `app/(marketing)/constructs/[slug]/page.tsx` | SoftwareApplication JSON-LD (enriched) |
| `app/(site)/about/page.tsx` | Organization JSON-LD, stats, richer meta |
| `convex/telegram.ts` | Telegram Bot API digest delivery (new) |
| `convex/analytics.ts` | Digest metrics + baseline queries (new) |
| `convex/crons.ts` | Daily digest cron at 14:00 UTC |
| `.env.example` | Umami + Telegram env var documentation |

## Test plan

- [ ] Verify Umami script loads on constructs.network (check network tab for `cloud.umami.is/script.js`)
- [ ] Verify JSON-LD renders on construct detail page (View Source → search `application/ld+json`)
- [ ] Verify about page shows live construct/skill counts
- [ ] Verify digest cron fires and sends to Telegram group (check Convex logs)
- [ ] Verify digest handles missing env vars gracefully (logs warning, doesn't crash)

### DIFF
--- a/apps/explorer/.env.example
+++ b/apps/explorer/.env.example
@@ -22,3 +22,12 @@ LINEAR_WEBHOOK_SECRET=
 # LINEAR_TEAM_INFRASTRUCTURE=<team-id>
 # Required in Convex dashboard for Discord alerting:
 # DISCORD_SIGNALS_WEBHOOK_URL=https://discord.com/api/webhooks/...
+
+# --- Analytics (cycle-051) ---
+# Umami Cloud — get website ID from https://cloud.umami.is
+NEXT_PUBLIC_UMAMI_WEBSITE_ID=
+# Required in Convex dashboard for Telegram digest:
+# TELEGRAM_BOT_TOKEN=<from @BotFather>
+# TELEGRAM_CHAT_ID=<group chat ID, negative number>
+# UMAMI_API_KEY=<from Umami Cloud settings>
+# UMAMI_WEBSITE_ID=<same as NEXT_PUBLIC_UMAMI_WEBSITE_ID>
--- a/apps/explorer/app/(marketing)/constructs/[slug]/page.tsx
+++ b/apps/explorer/app/(marketing)/constructs/[slug]/page.tsx
@@ -54,6 +54,16 @@ export default async function ConstructDetailPage({
       price: '0',
       priceCurrency: 'USD',
     },
+    author: {
+      '@type': 'Organization',
+      name: 'Constructs Network',
+      url: 'https://constructs.network',
+    },
+    url: `https://constructs.network/constructs/${slug}`,
+    ...(construct.version && { version: construct.version }),
+    ...(construct.verifiedAt
+      ? { datePublished: construct.verifiedAt }
+      : { datePublished: '2025-01-01' }),
   };
 
   const verificationBadge = (() => {
--- a/apps/explorer/app/(site)/about/page.tsx
+++ b/apps/explorer/app/(site)/about/page.tsx
@@ -7,7 +7,7 @@ import { fetchAllConstructs } from '@/lib/data/fetch-constructs';
 
 export const metadata: Metadata = {
   title: 'About | Constructs Network',
-  description: 'Skills for AI coding agents. Deploy, browse, compose.',
+  description: 'Skills for AI coding agents. 23 constructs, 150+ skills. Install a construct — your agent sees problems differently. Built by 0xHoneyJar.',
 };
 
 // ISR — revalidate hourly (matches homepage)
@@ -23,8 +23,46 @@ function RedactedSlot({ width = 'w-32' }: { width?: string }) {
   );
 }
 
+function AboutJsonLd({ constructCount, skillCount }: { constructCount: number; skillCount: number }) {
+  const jsonLd = {
+    '@context': 'https://schema.org',
+    '@graph': [
+      {
+        '@type': 'WebSite',
+        name: 'Constructs Network',
+        url: 'https://constructs.network',
+        description: `Skills for AI coding agents. ${constructCount} constructs, ${skillCount}+ skills.`,
+      },
+      {
+        '@type': 'Organization',
+        name: 'Constructs Network',
+        url: 'https://constructs.network',
+        foundingDate: '2026-02',
+        description: 'The open agent expertise network. Skills you install into your AI coding agent.',
+        sameAs: [
+          'https://x.com/zksoju',
+          'https://github.com/0xHoneyJar',
+        ],
+        parentOrganization: {
+          '@type': 'Organization',
+          name: '0xHoneyJar',
+          url: 'https://www.0xhoneyjar.xyz',
+        },
+      },
+    ],
+  };
+
+  return (
+    <script
+      type="application/ld+json"
+      dangerouslySetInnerHTML={{ __html: JSON.stringify(jsonLd) }}
+    />
+  );
+}
+
 export default async function AboutPage() {
   const allConstructs = await fetchAllConstructs();
+  const totalSkills = allConstructs.reduce((sum, c) => sum + (c.skillsCount || 0), 0);
 
   // Build a slug→logoWordmark map for constructs shown on this page
   const marks: Record<string, string | null> = {};
@@ -36,6 +74,7 @@ export default async function AboutPage() {
 
   return (
     <div className="mx-auto max-w-5xl px-4 py-16 sm:py-24">
+      <AboutJsonLd constructCount={allConstructs.length} skillCount={totalSkills} />
       <div className="space-y-32 sm:space-y-40">
 
         {/* Hero */}
@@ -50,6 +89,10 @@ export default async function AboutPage() {
               Skills you install into your AI coding agent.
               One command. Your agent sees problems differently.
             </p>
+            <div className="mt-6 font-mono text-sm text-bone-ghost">
+              {allConstructs.length} constructs · {totalSkills}+ skills · Open source · Built by{' '}
+              <a href="https://www.0xhoneyjar.xyz" className="text-bone-muted hover:text-bone-base transition-colors" target="_blank" rel="noopener noreferrer">0xHoneyJar</a>
+            </div>
           </section>
         </ScrollReveal>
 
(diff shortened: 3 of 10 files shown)


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/loa-constructs#173. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "FIND-001",
      "title": "Unbounded .collect() on potentially large tables",
      "severity": "HIGH",
      "category": "Performance",
      "file": "apps/explorer/convex/analytics.ts",
      "description": "Both `getDigestMetrics` and `getDailyBaseline` use `.collect()` on queries against `installEvents`, `signals`, and `healthObservations` tables. These queries fetch all matching documents into memory. As the application grows, a 24-hour or 7-day window could contain thousands or more records. Convex queries have execution limits, and unbounded collects can hit those limits or cause excessive memory usage. The `getDailyBaseline` query at line 75 collects all health observations in 7 days just to compute an average.",
      "suggestion": "Consider using aggregation patterns or adding `.take(limit)` with pagination. For the health average, consider sampling or maintaining a pre-computed rolling average. At minimum, add a reasonable upper bound like `.take(10000)` to prevent runaway queries.",
      "confidence": 0.75
    },
    {
      "id": "FIND-002",
      "title": "Baseline average divides by hardcoded 6 days, not actual days with data",
      "severity": "MEDIUM",
      "category": "Logic",
      "file": "apps/explorer/convex/analytics.ts",
      "description": "At line 90, the baseline calculation hardcodes `const days = 6` with the comment '7d window minus today'. However, the system may not have 6 full days of data (e.g., fresh deployment or after a data wipe). Dividing by 6 when there's only 1-2 days of data will artificially deflate the baseline, potentially suppressing anomaly detection. Additionally, the exclusion logic (7d minus last 24h) doesn't always produce exactly 6 days depending on when the cron runs.",
      "suggestion": "Calculate the actual number of distinct days in the data window, or at least track the earliest record timestamp to compute the true span. Fall back to `installs7d.length` as the average if insufficient history exists.",
      "confidence": 0.8
    },
    {
      "id": "FIND-003",
      "title": "Telegram bot token exposed in server logs on failure",
      "severity": "MEDIUM",
      "category": "Security",
      "file": "apps/explorer/convex/telegram.ts",
      "description": "The `sendMessage` function constructs the URL with the bot token embedded (`/bot${botToken}/sendMessage`). If the fetch fails with a network error (not a Telegram API error), the URL containing the token could appear in stack traces or error logs. The function only handles JSON response errors but not fetch-level exceptions.",
      "suggestion": "Wrap the fetch call in a try-catch block. In the catch handler, log a sanitized error message that does not include the URL. Consider storing the token reference rather than the value in any logged context.",
      "confidence": 0.6
    },
    {
      "id": "FIND-004",
      "title": "No error handling for fetch failure or non-JSON response in sendMessage",
      "severity": "MEDIUM",
      "category": "Reliability",
      "file": "apps/explorer/convex/telegram.ts:12-28",
      "description": "The `sendMessage` function does not check `response.ok` before calling `response.json()`. If the Telegram API returns a non-200 status with a non-JSON body, or if the network request fails entirely, this will throw an unhandled exception. Since this runs in a Convex action, the unhandled exception will cause the action to fail and potentially retry (depending on Convex retry policy), which could lead to duplicate messages if the first attempt actually succeeded but the response parsing failed.",
      "suggestion": "Add a try-catch around the fetch and JSON parsing. Check `response.ok` or `response.status` before parsing. Return a structured error on failure rather than letting exceptions propagate.",
      "confidence": 0.85
    },
    {
      "id": "FIND-005",
      "title": "Hardcoded founding date '2026-02' is in the future",
      "severity": "LOW",
      "category": "Correctness",
      "file": "apps/explorer/app/(site)/about/page.tsx:40",
      "description": "The structured data JSON-LD contains `foundingDate: '2026-02'`. Assuming the current date is in 2025, this date is in the future, which is semantically incorrect for a founding date and could confuse search engine parsers.",
      "suggestion": "Verify the correct founding date. If the organization was founded in 2025, update to '2025-02' or the actual date.",
      "confidence": 0.9
    },
    {
      "id": "FIND-006",
      "title": "Fallback datePublished '2025-01-01' is arbitrary",
      "severity": "LOW",
      "category": "SEO",
      "file": "apps/explorer/app/(marketing)/constructs/[slug]/page.tsx:63",
      "description": "When `construct.verifiedAt` is not available, the code falls back to `datePublished: '2025-01-01'`. This hardcoded date may be inaccurate for constructs published at different times and could mislead search engines about content freshness.",
      "suggestion": "Consider using a `createdAt` field if available, or omit `datePublished` entirely when the actual date is unknown rather than using an arbitrary fallback.",
      "confidence": 0.7
    },
    {
      "id": "FIND-007",
      "title": "Duplicate JSON-LD WebSite schema in layout and about page",
      "severity": "LOW",
      "category": "SEO",
      "file": "apps/explorer/app/layout.tsx:68",
      "description": "The root layout injects a `WebSite` JSON-LD schema (line 68-76), and the about page also injects a `WebSite` type within its `@graph` (about/page.tsx line 34-39). When visiting the about page, both JSON-LD blocks will be present, resulting in duplicate WebSite declarations that could confuse structured data parsers.",
      "suggestion": "Remove the WebSite schema from either the about page's `@graph` or the root layout to avoid duplication. Alternatively, only include the Organization type in the about page's JSON-LD.",
      "confidence": 0.85
    },
    {
      "id": "FIND-008",
      "title": "Hardcoded construct count '23' in meta description",
      "severity": "MEDIUM",
      "category": "Correctness",
      "file": "apps/explorer/app/(site)/about/page.tsx:10",
      "description": "The metadata description hardcodes '23 constructs, 150+ skills' but the page body dynamically computes `allConstructs.length` and `totalSkills`. The static metadata will become stale as constructs are added or removed, leading to a mismatch between the meta description and the page content.",
      "suggestion": "Either use `generateMetadata()` to dynamically set the description based on fetched data, or use a generic description that doesn't include specific counts.",
      "confidence": 0.9
    },
    {
      "id": "FIND-009",
      "title": "Script tag inside <head> managed by Next.js may cause hydration issues",
      "severity": "MEDIUM",
      "category": "Compatibility",
      "file": "apps/explorer/app/layout.tsx:83-89",
      "description": "Adding a `<head>` element directly in the root layout's `<html>` tag can conflict with Next.js App Router's built-in head management. Next.js App Router expects metadata to be managed via the `metadata` export or `<Metadata>` component. Manually adding `<head>` with `<Script>` inside it may lead to unexpected behavior or hydration mismatches, especially with the conditional rendering based on `process.env.NEXT_PUBLIC_UMAMI_WEBSITE_ID`.",
      "suggestion": "Move the Umami `<Script>` tag to the `<body>` instead (it uses `strategy='afterInteractive'` which works fine in body), or use Next.js metadata API for head management.",
      "confidence": 0.6
    },
    {
      "id": "FIND-010",
      "title": "Clean separation of analytics queries from digest delivery",
      "severity": "PRAISE",
      "category": "Architecture",
      "file": "apps/explorer/convex/analytics.ts",
      "description": "Good separation of concerns: the analytics queries (`getDigestMetrics`, `getDailyBaseline`) are in their own module and exposed as internal queries, while the Telegram delivery logic is in a separate file. This makes the metrics reusable for other consumers (e.g., a dashboard or different notification channel) without coupling to the delivery mechanism.",
      "suggestion": null
    },
    {
      "id": "FIND-011",
      "title": "XSS prevention in layout JSON-LD but not in about page",
      "severity": "MEDIUM",
      "category": "Security",
      "file": "apps/explorer/app/(site)/about/page.tsx:57",
      "description": "The root layout (layout.tsx:96) applies `.replace(/</g, '\\\\u003c')` to sanitize the JSON-LD output against XSS via `</script>` injection. However, the about page's `AboutJsonLd` component at line 57 uses `JSON.stringify(jsonLd)` directly without this sanitization. While the current data is static/numeric, if dynamic user-influenced data is ever included (e.g., construct names), this could become an XSS vector.",
      "suggestion": "Apply the same `.replace(/</g, '\\\\u003c')` sanitization in the about page's JSON-LD output for consistency and defense in depth.",
      "confidence": 0.7
    },
    {
      "id": "FIND-012",
      "title": "Index assumptions for installEvents and signals may not match schema",
      "severity": "MEDIUM",
      "category": "Correctness",
      "file": "apps/explorer/convex/analytics.ts:14-23",
      "description": "The queries use indexes `by_created` on `installEvents` and `by_timestamp` on `signals` with string-based ISO timestamp comparisons. This assumes these indexes exist in the Convex schema and that the indexed field stores ISO 8601 strings that sort lexicographically correctly. If the schema defines these indexes differently (e.g., on a numeric timestamp or different field name), the queries will fail at runtime.",
      "suggestion": "Verify that the Convex schema file defines these exact indexes on string-type timestamp fields. Consider adding a comment referencing the schema definition for maintainability.",
      "confidence": 0.5
    }
  ]
}
```


## Callouts

PR #185, opened 2026-03-24
fix: close construct sync lifecycle — skills discoverable after upgrade and session start

### LINKED ISSUE #120: Construct skill discovery breaks when skills are added post-install
## Problem

When a construct pack is installed via `/constructs`, the install script creates symlinks from `.claude/skills/{slug}` → `../constructs/packs/{pack}/skills/{slug}` for skill discovery. But this only happens **at install time** with whatever skills exist in the manifest at that moment.

When new skills are added to an installed construct — either by the pack author in a new version or by the repo developer extending the pack locally — **no symlinks are created**. The skills exist on disk but are invisible to Claude Code's skill discovery.

### Concrete example

Observer pack v1.6.0 manifest lists 16 skills. Only 4 have symlinks:

```
.claude/skills/observing-users  → ../constructs/packs/observer/skills/observing-users   ✓
.claude/skills/shaping-journeys → ../constructs/packs/observer/skills/shaping-journeys  ✓
.claude/skills/level-3-diagnostic → ../constructs/packs/observer/skills/level-3-diagnostic ✓
.claude/skills/importing-research → ../constructs/packs/observer/skills/importing-research ✓

# Missing — 12 skills with no symlinks:
generating-followups, ingesting-dms, daily-synthesis, batch-observing,
analyzing-gaps, filing-gaps, batch-filing-gaps, feedback-observing,
detecting-staleness, detecting-drift, refreshing-artifacts, snapshotting
```

These 12 skills work fine when invoked by reading SKILL.md directly, but `/follow-up` as a slash command fails with "Unknown skill" because the Skill tool can't find them.

## Root Cause

The construct lifecycle has **distribution** (versions, manifests, install) but lacks **development-in-place sync**. Constructs are designed to be portable packages that can be built, modified, and extended within any repo — but the install mechanism is fire-and-forget:

1. **No re-sync**: Adding a skill to `manifest.json` doesn't create its symlink
2. **No update hook**: `/update-loa` updates the framework but doesn't re-sync construct symlinks
3. **No local-dev awareness**: If a developer adds a skill to an installed pack (like we just did with the E5 follow-up rewrite), they have to manually create the symlink

### Option A: Manifest-driven sync command

```bash
/constructs sync [pack-name]
# Reads manifest.json, creates missing symlinks, removes stale ones
```

### Option B: Automatic sync on construct update

When a construct version bumps or a skill is added to the manifest, the update mechanism should diff the symlinks and create missing ones.

### Option C: Dynamic discovery (no symlinks needed)

Claude Code's skill discovery reads `.claude/skills/` for SKILL.md files. An alternative is to also scan `.claude/constructs/packs/*/skills/*/SKILL.md` at discovery time — eliminating the symlink indirection entirely.

## Broader Design Gap

Constructs are partially modeled after distributable packages (npm, crates) with versioning and a network. But unlike those systems:

- **No `postinstall` re-run on update** — symlinks created at v1.0.0 aren't refreshed at v1.6.0
- **No bidirectional development** — you can't easily extend a construct locally and contribute back
- **No divergence tracking** — if you modify a construct skill in-repo, there's no way to know it diverged from upstream

The construct network has distribution (releases, versions) but lacks the development-in-place workflow: modify construct in repo → sync → test → contribute back. This is the gap between "installable package" and "forkable repo with commit history."

## Affected Repos

- **loa-constructs**: Methodology/topology — constructs need lifecycle-aware install
- **loa**: Framework install mechanism — `scripts/install.sh` and skill discovery need sync support
- **midi-interface**: Discovered here — 12/16 observer skills invisible, manually fixed with `ln -s`

## Workaround

Manual symlink creation:
```bash
ln -s ../constructs/packs/observer/skills/generating-followups .claude/skills/generating-followups
```

### LINKED ISSUE #171: bug: installed packs go stale — no local-to-installed sync path
## Bug Report

**Source**: Discovered during deep research session (2026-03-14) when `dig-search.ts` from k-hole construct failed to find `.env` — the installed pack was months behind the canonical repo.

### Problem

Installed packs at `.claude/constructs/packs/{slug}/` are frozen snapshots with no automated update path from local canonical repos.

**Root cause chain:**
1. `do_install_pack_git()` clones with `--depth 1`, strips `.git/`, copies files
2. The installed pack is now a dead snapshot — no git history, no pull capability
3. `do_upgrade_pack()` exists but requires a **registry API roundtrip** — it checks the registry for a newer version, not the local filesystem
4. If the construct author updates `~/Documents/GitHub/construct-{slug}/` and pushes to GitHub but doesn't sync to the registry, the installed pack stays frozen
5. `/update-loa` updates the framework, not packs
6. No command exists for "sync this installed pack from the co-located canonical repo on my machine"

### Impact

**Specific failure encountered:** `construct-k-hole/scripts/dig-search.ts` canonical version had a proper `.env` directory walker (walks up to filesystem root). The installed version at `.claude/constructs/packs/k-hole/scripts/dig-search.ts` still had the old 2-directory walker that only checked the script dir and one level up — never finding `.env` at the project root (5 levels up from the pack install path).

This caused:
- `dig-search.ts` failing with "Missing GEMINI_API_KEY" despite the key being present in `.env`
- Research subagents falling back to manual `WebSearch`/`WebFetch` (15+ min vs 36 seconds)
- All `/dig` invocations silently broken for any project using installed (non-linked) packs

### Proposed Fix

One or more of:

1. **Local sync detection**: When a construct is installed and a co-located repo exists at `~/Documents/GitHub/construct-{slug}/`, detect version drift and offer sync
2. **`--local` flag for upgrade**: `constructs-install.sh upgrade k-hole --local ~/Documents/GitHub/construct-k-hole`
3. **Staleness warning at invoke time**: When a skill from an installed pack is invoked, check if the canonical repo (if co-located) has commits newer than the install timestamp
4. **Prefer symlinks for local dev**: The install script already handles symlinked packs differently (line 1477: "Pack is linked — use /construct-sync instead"). For developers who have the construct repo locally, symlink should be the default install mode

### Workaround

Manual copy: `cp construct-repo/scripts/*.ts .claude/constructs/packs/slug/scripts/`

### Environment

- loa-constructs v2.1.0+
- Affected: any installed (non-linked) pack where the canonical repo is updated independently
- The `do_upgrade_pack()` function at line 1462 of `constructs-install.sh` has the registry-based upgrade logic but no local-repo awareness

---

Filed by [Loa Framework](https://github.com/0xHoneyJar/loa) `/feedback` command

zkSoju:
## Summary

Closes the construct install→upgrade→sync lifecycle gap that caused skills to silently disappear after pack evolution. Three root causes fixed:

- **`do_upgrade_pack()` never re-linked skills** — 3-way merge completed but new skills from upgrades stayed invisible to Claude Code's skill discovery
- **`sync-packs.sh` (SessionStart hook) skipped `.claude/skills/`** — synced pack-level and command symlinks but never touched skill symlinks
- **No manual re-link path existed** — `link-commands` had no `link-skills` counterpart

## Changes

| File | What |
|------|------|
| `.claude/scripts/constructs-install.sh` | Phase 7 in upgrade, `do_link_skills()`, `do_link_all()`, `clean_stale_skill_symlinks()`, `link-skills` + `link` CLI subcommands, boundary-safe `validate_symlink_target()` |
| `scripts/sync-packs.sh` | Skill symlink sync section, stale cleanup via `find -type l`, canonical regex validation, `LC_ALL=C` deterministic ordering |

## Security hardening (from GPT-5.2 cross-model review)

- Boundary-safe prefix matching with slash delimiter (prevents sibling-path bypass like `/x/constructs-evil` matching `/x/constructs*`)
- Canonical regex for trusted symlink targets: `^../constructs/packs/[^/]+/skills/[^/]+$`
- `find -type l` for stale cleanup (glob `*/` misses dangling symlinks)
- `HOME`-scoped trust boundary on `LOA_GLOBAL_STORE` env var
- Defensive numeric coercion on `symlink_pack_skills` output

## Verification

| Check | Result |
|-------|--------|
| First sync run | 107 skills linked, 6 stale removed |
| Second run (idempotency) | 0 linked, 160 skipped, 0 stale |
| Timing (24 packs, 160 skills) | 1.27s |
| `link-skills observer` | 24/24 linked |
| `link all` | 160 skills, 66 commands |
| `constructs-health.sh` | No regression (143/192 pre-existing) |
| `bash -n` syntax check | Both files pass |
| Collision handling | 2 collisions (`research`, `configure-project`) — first-wins-stable |

## Test plan

- [ ] Verify `sync-packs.sh --json` shows `skills_linked: 0` on clean repo (idempotent)
- [ ] Install a new pack, verify skills appear in `.claude/skills/`
- [ ] Remove a skill from a pack, verify stale symlink is cleaned on next sync
- [ ] Run `constructs-install.sh link all` — all skills and commands linked
- [ ] Verify SessionStart hook completes in < 2 seconds

Closes #120
Closes #171

### DIFF
--- a/.claude/scripts/constructs-install.sh
+++ b/.claude/scripts/constructs-install.sh
@@ -156,9 +156,36 @@ validate_symlink_target() {
                 constructs_abs=$(readlink -f "$constructs_dir" 2>/dev/null || realpath "$constructs_dir" 2>/dev/null || (cd "$constructs_dir" 2>/dev/null && pwd -P) || echo "")
             fi
 
-            # Verify resolved path is within constructs
-            if [[ -n "$constructs_abs" ]] && [[ "$resolved_target" != "$constructs_abs"* ]]; then
-                print_warning "Symlink resolves outside constructs: $resolved_target"
+            # Verify resolved path is within constructs (local or global store)
+            # Trust boundary: only accept global store under $HOME to prevent
+            # env var injection from widening the validation boundary
+            local global_store="${LOA_GLOBAL_STORE:-${HOME}/.loa/constructs}"
+            local global_abs=""
+            if [[ -d "$global_store" ]] && [[ "$global_store" == "${HOME}/"* ]]; then
+                if command -v get_canonical_path &>/dev/null; then
+                    global_abs=$(get_canonical_path "$global_store")
+                else
+                    global_abs=$(readlink -f "$global_store" 2>/dev/null || realpath "$global_store" 2>/dev/null || (cd "$global_store" 2>/dev/null && pwd -P) || echo "")
+                fi
+            fi
+
+            # Normalize trailing slashes for boundary-safe prefix checks
+            constructs_abs="${constructs_abs%/}"
+            global_abs="${global_abs%/}"
+
+            local in_local="false"
+            local in_global="false"
+
+            if [[ -n "$constructs_abs" ]] && { [[ "$resolved_target" == "$constructs_abs" ]] || [[ "$resolved_target" == "$constructs_abs/"* ]]; }; then
+                in_local="true"
+            fi
+
+            if [[ -n "$global_abs" ]] && { [[ "$resolved_target" == "$global_abs" ]] || [[ "$resolved_target" == "$global_abs/"* ]]; }; then
+                in_global="true"
+            fi
+
+            if [[ "$in_local" != "true" ]] && [[ "$in_global" != "true" ]]; then
+                print_warning "Symlink resolves outside constructs/global store: $resolved_target"
                 return 1
             fi
         fi
@@ -1627,6 +1654,13 @@ do_upgrade_pack() {
     # Phase 6: Update metadata
     update_pack_meta "$pack_slug" "$pack_dir"
 
+    # Phase 7: Re-link skills and commands (new skills from upgrade become discoverable)
+    local skills_linked commands_linked
+    skills_linked=$(symlink_pack_skills "$pack_slug" 2>/dev/null)
+    skills_linked=${skills_linked:-0}
+    commands_linked=$(symlink_pack_commands "$pack_slug" 2>/dev/null)
+    commands_linked=${commands_linked:-0}
+
     # Cleanup
     rm -rf "$tmp_new" "$backup_dir"
 
@@ -1635,6 +1669,9 @@ do_upgrade_pack() {
     echo "  Kept local:   $merged files"
     if [[ $conflicts -gt 0 ]]; then
         print_warning "  Conflicts:    $conflicts files (check .upstream files)"
+    fi
+    if [[ "$skills_linked" -gt 0 ]] || [[ "$commands_linked" -gt 0 ]]; then
+        echo "  Skills linked: $skills_linked, Commands linked: $commands_linked"
     fi
     echo ""
     print_success "Pack '$pack_slug' upgraded to $latest_version"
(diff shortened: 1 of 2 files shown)


zkSoju (commented):
## Summary

Analytical review of 0xHoneyJar/loa-constructs#185. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "FIND-001",
      "title": "Using `ls` for directory iteration instead of glob expansion",
      "severity": "MEDIUM",
      "category": "Correctness",
      "file": "scripts/sync-packs.sh",
      "description": "Line 200 uses `$(LC_ALL=C ls -d \"$LOCAL_PACKS\"/*/ 2>/dev/null)` in a for loop. Parsing `ls` output is fragile — directory names containing whitespace, newlines, or glob characters will break iteration. The `LC_ALL=C` comment says it's for deterministic order, but the real risk is word-splitting on spaces in paths.",
      "suggestion": "Use a sorted glob pattern instead: `for pack_dir in \"$LOCAL_PACKS\"/*/; do` combined with sorting if deterministic order is truly needed (e.g., pipe through `sort -z` with null delimiters, or use `mapfile` with `< <(printf '%s\\0' \"$LOCAL_PACKS\"/*/ | LC_ALL=C sort -z)`). This avoids parsing `ls` output entirely.",
      "confidence": 0.75
    },
    {
      "id": "FIND-002",
      "title": "Stale symlink cleanup uses `-type l` which may not catch all dangling symlinks on all platforms",
      "severity": "LOW",
      "category": "Portability",
      "file": "scripts/sync-packs.sh",
      "description": "On line 248, `find ... -type l` is used. On most modern systems (GNU find, BSD find on macOS), `-type l` matches symbolic links including dangling ones. However, the comment says 'Uses find -type l to catch dangling symlinks' — this is correct on GNU/BSD find but worth noting the script should verify this works on the target platforms.",
      "suggestion": "This is likely fine for the target environments (macOS + Linux). No action needed unless exotic platforms are supported.",
      "confidence": 0.3
    },
    {
      "id": "FIND-003",
      "title": "Symlink target validation resolves relative to LOCAL_SKILLS but traversal could escape",
      "severity": "LOW",
      "category": "Security",
      "file": "scripts/sync-packs.sh",
      "description": "Line 222 creates symlinks with `relative_path=\"../constructs/packs/${slug}/skills/${skill_name}\"` where `slug` and `skill_name` come from `basename` calls. `basename` strips directory components, so path traversal via `slug` or `skill_name` is prevented. The regex check on line 233 (`^[.][.]/constructs/packs/[^/]+/skills/[^/]+$`) also validates the pattern. This looks well-defended.",
      "suggestion": "No change needed — the use of `basename` and the regex validation are sufficient safeguards.",
      "confidence": 0.1
    },
    {
      "id": "FIND-004",
      "title": "First-wins collision semantics depend on `ls` sort order, not explicit priority",
      "severity": "MEDIUM",
      "category": "Correctness",
      "file": "scripts/sync-packs.sh",
      "description": "If two packs provide a skill with the same name, the first one encountered wins (line 230-235 keeps existing valid symlinks). The iteration order depends on `LC_ALL=C ls` output, which is lexicographic by directory name. This is deterministic but implicit — there's no explicit priority mechanism. A user adding a pack named `aaa-override` would silently win over `zzz-original` with no warning.",
      "suggestion": "Consider logging a warning when a skill name collision is detected across packs, so users are aware which pack's skill is being used. e.g., `echo \"warning: skill '${skill_name}' from pack '${slug}' shadowed by existing link\" >&2`",
      "confidence": 0.65
    },
    {
      "id": "FIND-005",
      "title": "Clean symlink lifecycle with framework built-in protection",
      "severity": "PRAISE",
      "category": "Design",
      "file": "scripts/sync-packs.sh",
      "description": "The pattern of checking `[[ -e \"$local_skill\" ]] && [[ ! -L \"$local_skill\" ]]` to protect real directories/files from being overwritten by pack symlinks is a good defensive design. The stale symlink cleanup also correctly restricts itself to only canonical construct links via regex, avoiding accidental deletion of user-created symlinks.",
      "suggestion": "None — this is well-designed.",
      "confidence": 0.95
    },
    {
      "id": "FIND-006",
      "title": "Trailing slash glob pattern may not match if skills directory is empty",
      "severity": "LOW",
      "category": "Correctness",
      "file": "scripts/sync-packs.sh",
      "description": "Line 208: `for skill_path in \"$skills_dir\"/*/;` — if the skills directory exists but is empty, the glob won't expand and the loop body will execute once with the literal pattern. The guard `[[ -d \"$skill_path\" ]] || continue` on line 209 handles this correctly.",
      "suggestion": "Already handled by the `-d` guard. No action needed.",
      "confidence": 0.9
    }
  ]
}
```


## Callouts

PR #223, opened 2026-05-03
feat(shared): substrate-construct manifest schema + 11 tests

zkSoju:
## Summary

- Extends \`packManifestSchema\` (Zod) with the **substrate-construct** pack type — executable Effect programs that take typed input and yield typed output via the Effect Requirements channel.
- Adds 5 new Zod schemas: \`substrateRuntimeSchema\`, \`substrateExecutableSchema\`, \`substrateRequirementSchema\`, \`substrateStreamEntrySchema\`, \`substrateStreamsSchema\`.
- Adds 4 new optional fields to \`packManifestSchema\` (\`runtime\`, \`executable\`, \`requirements\`, \`streams\`).
- Adds \`superRefine\` conditional: when \`type === 'substrate-construct'\`, both \`executable\` and \`runtime\` are required.
- Adds 11 new tests in \`pack-manifest.test.ts\` (all 73 tests pass).

## Why

Cycle 2026-05-03 substrate-integration session surfaced a missing convention: today's \`construct-*\` repos are all skill-packs (markdown + slash commands loaded by Claude Code). The substrate-mental-model doctrine described an executable substrate (Effect programs that ride NATS/Kafka, run in Finn sandboxes, declare typed Requirements) but the pack type didn't exist. This PR establishes the manifest TYPES (Zod) for that type — paired with the JSON Schema in [construct-base#12](https://github.com/0xHoneyJar/construct-base/pull/12).

Per operator framing: \"the types and schemas around this can live with the loa-constructs (which holds schemas for the network just like loa-hounfour)\". This PR is loa-constructs's substrate-construct shape; loa-hounfour holds the agent-network shapes.

First instance riding this schema: [0xHoneyJar/construct-lore-essay-grader](https://github.com/0xHoneyJar/construct-lore-essay-grader) — a BORGES-personified subjective essay grader.

## What's in the diff

\`\`\`
packages/shared/src/validation.ts          (+~150 lines)
packages/shared/src/__tests__/pack-manifest.test.ts  (+~140 lines, 11 new tests)
\`\`\`

5 new Zod schemas:
- \`substrateRuntimeSchema\`: \`{ engine, engine_version, node_version }\`
- \`substrateExecutableSchema\`: \`{ entry, export, protocol: { input, output } }\`
- \`substrateRequirementSchema\`: \`{ tag, contract, description }\` (Effect Tag declarations)
- \`substrateStreamEntrySchema\`: \`{ subject, schema, narrows_to/from }\` (NATS/Kafka topic shape)
- \`substrateStreamsSchema\`: \`{ reads, writes }\` (arrays of stream entries)

5 new TypeScript types exported alongside.

Conditional validation via \`superRefine\` (Zod's recommended pattern for cross-field invariants):
\`\`\`ts
if (manifest.type === 'substrate-construct') {
  if (!manifest.executable) ctx.addIssue({ ... })
  if (!manifest.runtime) ctx.addIssue({ ... })
}
\`\`\`

## Companion

- [construct-base#12](https://github.com/0xHoneyJar/construct-base/pull/12) — matching JSON Schema for ajv validation in CI
- [construct-lore-essay-grader](https://github.com/0xHoneyJar/construct-lore-essay-grader) — first instance riding the new shape

## Test plan

- [x] All 73 tests pass (62 existing + 11 new)
- [x] Backward-compat verified: skill-pack manifests still validate
- [x] Forward shape verified: substrate-construct manifest with all 4 new fields validates
- [x] Conditional rejection verified: substrate-construct missing executable OR runtime is rejected
- [x] Additive shape verified: skill-pack with executable also validates (executable not exclusive to substrate-construct)
- [ ] Operator review the field shapes (especially streams.subject convention) before merge

## Doctrine

\`~/vault/wiki/concepts/substrate-mental-model-for-product-builders.md\`

### DIFF
--- a/packages/shared/src/__tests__/pack-manifest.test.ts
+++ b/packages/shared/src/__tests__/pack-manifest.test.ts
@@ -886,10 +886,112 @@ describe('Substrate-Construct: type + executable + runtime + requirements + stre
     expect(result.success).toBe(false);
   });
 
-  it('rejects streams entry without subject', () => {
+  it('rejects streams object entry without subject', () => {
     const manifest = {
       ...SUBSTRATE_MANIFEST,
       streams: { reads: [{ schema: 'foo' }] },
+    };
+    const result = packManifestSchema.safeParse(manifest);
+    expect(result.success).toBe(false);
+  });
+
+  // ── Bridgebuilder F1 backward-compat fix (cycle 2026-05-03) ──────────
+
+  it('accepts streams cycle-002 string-shape (Intent, Verdict, Artifact)', () => {
+    // construct-creator and any future skill-pack riding the cycle-002
+    // typed-streams primitive use bare strings naming the stream type.
+    // The substrate-construct's substrateStreamEntrySchema must admit BOTH
+    // shapes (z.union of string OR object) — verified here.
+    const manifest = {
+      name: 'Construct Creator',
+      slug: 'construct-creator',
+      version: '1.0.0',
+      description: 'Skill-pack that emits typed streams (cycle-002 convention)',
+      type: 'skill-pack' as const,
+      skills: [{ slug: 'creating-constructs', path: 'skills/creating-constructs/' }],
+      streams: {
+        reads: ['Intent', 'Operator-Model'],
+        writes: ['Verdict', 'Artifact', 'Signal'],
+      },
+    };
+    const result = packManifestSchema.safeParse(manifest);
+    expect(result.success).toBe(true);
+  });
+
+  it('accepts streams mixed shape (string + object)', () => {
+    // Forward-compat — a pack might bridge cycle-002 typed-streams AND
+    // substrate-construct Kafka subjects in the same manifest. Mixed array
+    // must pass.
+    const manifest = {
+      ...SUBSTRATE_MANIFEST,
+      streams: {
+        reads: [
+          'Intent',
+          { subject: 'agent.lore-essay.submission', schema: '@freeside-quests/protocol#SubstrateStepSubmission' },
+        ],
+        writes: ['Verdict'],
+      },
+    };
+    const result = packManifestSchema.safeParse(manifest);
+    expect(result.success).toBe(true);
+  });
+
+  // ── Bridgebuilder F1 + F2 superRefine tightening (cycle 2026-05-03) ──
+
+  it('rejects substrate-construct with empty runtime: {} (engine missing)', () => {
+    // Previously passed because runtime was just-non-undefined. The Bridgebuilder
+    // F1 finding caught this: runtime.engine is the load-bearing field that
+    // tells the runtime layer which engine to spawn. Empty object defeats
+    // dispatch.
+    const manifest = { ...SUBSTRATE_MANIFEST, runtime: {} };
+    const result = packManifestSchema.safeParse(manifest);
+    expect(result.success).toBe(false);
+    if (!result.success) {
+      const issues = result.error.issues.map((i) => i.message);
+      expect(issues.some((m) => m.includes('runtime.engine'))).toBe(true);
+    }
+  });
+
+  it('rejects substrate-construct executable without protocol.input', () => {
+    // Doctrine emphasizes typed-input → typed-output; an executable without
+    // protocol refs is opaque. Bridgebuilder F2 finding.
+    const manifest = {
+      ...SUBSTRATE_MANIFEST,
+      executable: {
+        entry: 'src/index.ts',
+        export: 'doThing',
+        protocol: { output: 'src/protocol.ts#Output' }, // input missing
+      },
+    };
+    const result = packManifestSchema.safeParse(manifest);
+    expect(result.success).toBe(false);
+    if (!result.success) {
+      const issues = result.error.issues.map((i) => i.message);
+      expect(issues.some((m) => m.includes('executable.protocol.input'))).toBe(true);
+    }
+  });
+
+  it('rejects substrate-construct executable without protocol.output', () => {
+    const manifest = {
+      ...SUBSTRATE_MANIFEST,
+      executable: {
+        entry: 'src/index.ts',
+        export: 'doThing',
+        protocol: { input: 'src/protocol.ts#Input' }, // output missing
+      },
+    };
+    const result = packManifestSchema.safeParse(manifest);
+    expect(result.success).toBe(false);
+    if (!result.success) {
+      const issues = result.error.issues.map((i) => i.message);
+      expect(issues.some((m) => m.includes('executable.protocol.output'))).toBe(true);
+    }
+  });
+
+  it('rejects substrate-construct executable without any protocol declaration', () => {
+    const manifest = {
+      ...SUBSTRATE_MANIFEST,
+      executable: { entry: 'src/index.ts', export: 'doThing' }, // no protocol
     };
     const result = packManifestSchema.safeParse(manifest);
     expect(result.success).toBe(false);
--- a/packages/shared/src/__tests__/pack-manifest.test.ts
+++ b/packages/shared/src/__tests__/pack-manifest.test.ts
@@ -762,3 +762,136 @@ describe('Construct Lifecycle: meta_probe drift reconciliation', () => {
     expect(result.success).toBe(true);
   });
 });
+
+// ── Substrate-Construct schema (cycle 2026-05-03 substrate-integration) ──
+
+describe('Substrate-Construct: type + executable + runtime + requirements + streams', () => {
+  const SUBSTRATE_MANIFEST = {
+    name: 'Lore Essay Grader',
+    slug: 'lore-essay-grader',
+    version: '0.1.0',
+    description: 'BORGES-personified subjective essay grader for substrate-graded activity steps',
+    type: 'substrate-construct' as const,
+    runtime: {
+      engine: 'effect-ts' as const,
+      engine_version: '^3.10.0',
+      node_version: '>=20.0.0',
+    },
+    executable: {
+      entry: 'src/index.ts',
+      export: 'gradeLoreEssay',
+      protocol: {
+        input: 'src/protocol.ts#LoreEssayInput',
+        output: 'src/protocol.ts#LoreEssayOutput',
+      },
+    },
+    requirements: [
+      { tag: 'ModelRunner', contract: 'src/grader.ts#ModelRunner', description: 'LLM invocation port' },
+      { tag: 'Logger', contract: 'effect/Logger' },
+    ],
+    streams: {
+      reads: [
+        {
+          subject: 'agent.lore-essay.submission',
+          schema: '@freeside-quests/protocol#SubstrateStepSubmission',
+          narrows_to: 'src/protocol.ts#LoreEssayInput',
+        },
+      ],
+      writes: [
+        {
+          subject: 'agent.lore-essay.verdict',
+          schema: '@freeside-quests/protocol#SubstrateStepVerdict',
+          from: 'src/protocol.ts#LoreEssayOutput',
+        },
+      ],
+    },
+  };
+
+  it('accepts a fully-formed substrate-construct manifest', () => {
+    const result = packManifestSchema.safeParse(SUBSTRATE_MANIFEST);
+    expect(result.success).toBe(true);
+  });
+
+  it('accepts substrate-construct without skills (skills optional for this type)', () => {
+    const manifest = { ...SUBSTRATE_MANIFEST };
+    const result = packManifestSchema.safeParse(manifest);
+    expect(result.success).toBe(true);
+  });
+
+  it('accepts skill-pack without runtime/executable (backward compat)', () => {
+    const manifest = {
+      name: 'Skill Pack',
+      slug: 'skill-pack-test',
+      version: '1.0.0',
+      description: 'A normal skill-pack manifest',
+      type: 'skill-pack' as const,
+      skills: [{ slug: 'do-thing', path: 'skills/do-thing/' }],
+    };
+    const result = packManifestSchema.safeParse(manifest);
+    expect(result.success).toBe(true);
+  });
+
+  it('rejects substrate-construct missing executable', () => {
+    const { executable, ...manifest } = SUBSTRATE_MANIFEST;
+    const result = packManifestSchema.safeParse(manifest);
+    expect(result.success).toBe(false);
+    if (!result.success) {
+      const issues = result.error.issues.map((i) => i.message);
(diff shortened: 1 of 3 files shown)


zkSoju (commented):
## Bridgebuilder Review · BEAUVOIR · PR #223

A clean, contained extension that earns its merge. The Zod additions land where they should, the conditional refinement uses `superRefine` correctly, and the 11 tests document both the happy path and the rejection contract. The doctrinal seam between `skill-pack` (markdown · Claude Code loads it) and `substrate-construct` (executable Effect program · runtime layer invokes it) is now expressible in the manifest — a real upgrade for the network's shape.

Most findings below are **about what's permissive that perhaps shouldn't be** — schema strictness questions where the additive ethos sometimes invites silent under-specification. The conditional refinement enforces field *presence* but not field *completeness*, and that gap will matter the moment the runtime layer starts injecting Tags. None of this is a blocker; it's all "before the third instance ships, decide which knobs you want."

---

### Findings

**F1 · `runtime: {}` passes validation despite being semantically empty** [Quality · MEDIUM]
*File: `packages/shared/src/validation.ts:342-346, 545-563`*

`substrateRuntimeSchema` declares `engine`, `engine_version`, and `node_version` all `.optional()`. The `superRefine` only checks `if (!manifest.runtime)` — a manifest declaring `runtime: {}` (empty object) satisfies the conditional and parses successfully, even though the runtime layer cannot decide which engine to invoke. The PR description even says "tells the runtime layer what engine + version + node target the executable expects" — empty object tells it nothing.

**FAANG parallel:** Kubernetes `PodSpec.containers[].resources` had this exact shape for years — `requests: {}` was valid YAML but useless to the scheduler. The fix landed in 1.16+ when the admission controller rejected empty resource specs for guaranteed-tier pods. Same pattern: schema *shape* validation isn't enough when the *contents* carry semantic load.

**Metaphor:** A passport with the photo page intact but every field blank. The binding is correct; the bearer is unidentifiable.

**Recommendation:** Either make `engine` required at the schema level (`z.enum([...])` without `.optional()`) OR extend the `superRefine` to require at least `runtime.engine` when `type === 'substrate-construct'`. The former is cleaner because the type narrowing flows to consumers; the latter preserves additive backward-compat if you're worried about not-yet-imagined runtime shapes.

**Decision trail:** The doctrine page says substrate-constructs are *executable Effect programs* — the executable invocation needs to know what engine to spawn. Empty-runtime is a bug class, not a feature.

---

**F2 · `executable.protocol.input/output` are optional, undermining the typed-IO doctrine** [Quality · MEDIUM]
*File: `packages/shared/src/validation.ts:359-362`*

The PR description and the cycle 2026-05-03 substrate-integration framing both emphasize "executable Effect programs that take typed input and yield typed output via the Effect Requirements channel." But `substrateExecutableSchema.protocol` makes both `input` and `output` `.optional()`, AND the entire `protocol` object is `.optional()`. A substrate-construct can declare `executable: { entry, export }` and pass — no protocol references at all.

**FAANG parallel:** gRPC's `.proto` files require `rpc Foo(InputMessage) returns (OutputMessage)` — both sides typed, no exceptions. The whole point of code-generation is that the contract is mechanically discoverable. When the IDL allows untyped methods (early Thrift had this escape hatch), the entire downstream tooling pipeline becomes "best effort" instead of "guaranteed."

**Metaphor:** A function signature that names itself but doesn't declare its parameters. The caller has to read the body to know what to pass.

**Recommendation:** When `type === 'substrate-construct'`, require `executable.protocol.input` and `executable.protocol.output` via `superRefine`. This catches under-specified manifests at registry-publish time, before they hit the runtime injection layer where the failure mode is "construct runs but produces unparseable output."

**Decision trail:** If the typed-IO doctrine is real, the schema should enforce it for the type that claims to embody it. Otherwise the convention drifts into "documentation by name."

---

**F3 · `streams.{reads,writes}` entry accepts both `narrows_to` AND `from` on either side** [Quality · LOW]
*File: `packages/shared/src/validation.ts:393-398`*

`substrateStreamEntrySchema` has both `narrows_to` (read-side per the comment) and `from` (write-side per the comment) as sibling-optional fields. A read entry with `from: 'src/protocol.ts#X'` validates fine. A write entry with `narrows_to: 'src/protocol.ts#Y'` validates fine. Both fields on the same entry validates fine. The doctrine in the schema docstring (`narrows_to` for reads, `from` for writes) is not mechanically enforced.

**FAANG parallel:** Kafka's Avro schema registry historically allowed both `key` and `value` schemas on either side of a topic — operators kept setting them on the wrong field and discovering at runtime their consumers couldn't deserialize. Confluent's resolution: a single `direction`-tagged union shape per side.

**Metaphor:** A USB-C connector that's physically reversible — a feature for cables, a bug for protocol semantics.

**Recommendation:** Two cleaner options. (a) Split the schema into `substrateStreamReadEntrySchema` (with `narrows_to`, no `from`) and `substrateStreamWriteEntrySchema` (with `from`, no `narrows_to`), then update `substrateStreamsSchema.reads` and `.writes` to use the directional shape. (b) Keep the unified shape but add a `superRefine` on `substrateStreamsSchema` that validates entries-by-position. (a) is the typesafe choice; (b) is the additive choice. Defer to whichever the JSON Schema sibling in construct-base#12 picked.

**Decision trail:** The naming asymmetry (`narrows_to` vs `from`) only earns its keep if the directional invariant holds. Today it's a comment, not a contract.

---

**F4 · Hand-written `PackManifest` interface in `types.ts` drifts from Zod schema** [Quality · MEDIUM]
*File: `packages/shared/src/types.ts:348-349` (drift relative to `validation.ts:292-298, 510-513`)*

`packages/shared/src/types.ts:349` declares `type?: 'skill-pack' | 'tool-pack' | 'codex' | 'template'` — missing `'substrate-construct'`. The same interface is missing `runtime`, `executable`, `requirements`, and `streams` fields entirely. Two consumers in this repo import `PackManifest` from `@loa-constructs/shared` (not the inferred `ValidatedPackManifest`):
- `packages/loa-registry/src/commands/pack-list.ts:10`
- `packages/loa-registry/src/commands/pack-update.ts:10`

Today neither narrows on `type` so nothing breaks at compile time. But the next consumer that writes `if (manifest.type === 'substrate-construct')` will get a TypeScript error: the literal isn't in the union. And anyone reading `manifest.executable?.entry` will get `Property 'executable' does not exist on type 'PackManifest'`.

**FAANG parallel:** This is the duplicate-source-of-truth bug class that drove the Stripe v3 → v4 SDK overhaul — they had hand-typed interfaces alongside OpenAPI schemas, and every API addition required two PRs that drifted in practice. The Stripe answer was full codegen from OpenAPI; the modern Zod answer is `z.infer<typeof packManifestSchema>` and delete the hand-written interface.

**Metaphor:** Two clocks in the same room, one mechanical and one quartz. They start aligned. They never stay aligned.

**Recommendation:** In a follow-up (not this PR — scope discipline), either (a) regenerate `PackManifest` from `z.infer<typeof packManifestSchema>` and re-export, OR (b) at minimum add the four substrate fields and `'substrate-construct'` to the type union in `types.ts`. The current PR can land as-is; the drift is latent. File a tracking issue against this PR before merge.

**Decision trail:** The Zod schema is the source of truth (validation runs against it). The hand-written interface is the consumer-facing type. They have to agree or one of them is lying.

---

**F5 · `.passthrough()` everywhere is principled but worth re-examining for substrate runtime** [Security/Quality · LOW]
*File: `packages/shared/src/validation.ts:346, 363, 379, 398, 410, 545`*

Every new substrate schema ends in `.passthrough()`. For the existing `packManifestSchema`, passthrough is the right call (preserves forward-compat for fields registered in the JSON Schema sibling). For the substrate-specific shapes, it's worth a moment of thought: these manifests will (per the PR description) drive runtime Tag injection in `loa-hounfour`. If a malicious or buggy pack ships `requirements: [{ tag: 'ModelRunner', __proto__: 'evil' }]`, the unknown key survives through validation. Today's consumers don't iterate, so it's safe; future consumers might.

**FAANG parallel:** This is the prototype-pollution attack vector that bit Lodash 4.17.x and most JSON-config-driven systems. The fix isn't `.strict()` everywhere — it's `.strict()` on the inner shapes whose keys flow into `Object.keys` / `for...in` iteration in code paths that touch object identity.

**Metaphor:** A hotel front desk that records the names you give but quietly notes the ones you didn't. Fine for memory; concerning if the notes get acted on.

**Recommendation:** Defer. But document the choice in the schema docstring ("passthrough is intentional for forward-compat — runtime consumers MUST treat unknown keys as data, not as code"), and consider `.strict()` on `substrateRequirementSchema` specifically before the first runtime layer ships in `loa-hounfour`. For now, status quo is fine — the attack surface is theoretical until the runtime exists.

**Decision trail:** Passthrough is the framework's house style. Don't change the house style for one schema. Do consider whether the runtime-injection surface deserves the exception.

---

**F6 · Test "accepts substrate-construct without skills" is identical to the prior test** [Test Coverage · LOW]
*File: `packages/shared/src/__tests__/pack-manifest.test.ts:815-819`*

```ts
it('accepts substrate-construct without skills (skills optional for this type)', () => {
  const manifest = { ...SUBSTRATE_MANIFEST };
  const result = packManifestSchema.safeParse(manifest);
  expect(result.success).toBe(true);
});
```

`SUBSTRATE_MANIFEST` doesn't define a `skills` field, and this test's manifest is `{ ...SUBSTRATE_MANIFEST }` — a structural identical copy. The test name suggests it exercises "skills omitted" as a distinct case from the prior test, but `safeParse` sees the same object shape both times. Either delete this test (the prior one already covers it) OR change it to explicitly assert `skills: []` works AND/OR `skills: undefined` works AND/OR a substrate-construct WITH skills (mixed type) works. The last case (mixed) is the genuinely interesting one given the design says skill-pack and substrate-construct are different archetypes.

**FAANG parallel:** Google's testing-blog-famous "tests that test the tests" — duplicate green tests give false coverage signal. Coverage tooling counts the assertions; reviewers count the distinct shapes exercised.

**Metaphor:** Pressing the same elevator button twice and counting it as two passengers.

**Recommendation:** Either remove (lowest cost) or expand into three sub-cases: (a) `skills: []`, (b) `skills` field absent, (c) substrate-construct WITH skills (does the doctrine forbid this? if so, add `superRefine` enforcement and a rejection test).

**Decision trail:** This is the kind of finding that earns its keep at the third or fourth instance. Today the schema is small enough that the test reads as documentation; once 5+ packs ride it, the suite earns rigor.

---

**F7 · Missing test: substrate-construct with `runtime.engine` set to a non-enum value** [Test Coverage · LOW]
*File: `packages/shared/src/__tests__/pack-manifest.test.ts` (no test exists)*

The 11 new tests cover field-presence (rejects-missing-X) but not field-content. There's no test that says "if you set `runtime.engine: 'deno'` (not in the enum), validation rejects." Same for `executable.export: ''` (empty string after `.min(1)`), `requirements[].tag: ''`, `streams.reads[].subject: ''`. The schema declares these constraints; the tests don't lock them in.

**FAANG parallel:** Property-based testing (fast-check, hypothesis) was invented for this exact gap — example-based tests document the happy path and the obvious rejection paths but miss the boundary conditions where regressions silently land. The Stripe SDK test suite intentionally has 5+ rejection tests per accept test for this reason.

**Metaphor:** A lock that's tested by the right key and one wrong key. The other 999 wrong keys are hoped against.

**Recommendation:** Add 4-6 boundary tests. Suggested:
- `engine: 'deno'` rejects (not in enum)
- `executable.entry: ''` rejects (`.min(1)`)
- `requirements[].tag: ''` rejects (`.min(1)`)
- `streams.reads[].subject: ''` rejects (`.min(1)`)
- `requirements` array of length 21 rejects (`.max(20)`)
- `streams.reads` array of length 21 rejects (`.max(20)`)

This is a 15-line PR follow-up, not a blocker. Land it before the construct-lore-essay-grader instance ships so the contract is durable.

**Decision trail:** The schema bounds are documented in the source. Tests should turn that documentation into enforced contract.

---

### Positive Callouts

**P1 · The conditional refinement is the right shape, in the right place.** `superRefine` on `packManifestSchema` (not on a separate schema imported at validation-call sites) means the conditional invariant lives with the union it constrains. Cross-field conditionals are the most error-prone Zod pattern; this PR uses the canonical form (custom-coded `addIssue` with `path: [field]`), gets the path right (so `result.error.issues` surfaces the field name to consumers), and tests both the missing-`executable` and missing-`runtime` paths in isolation. This is textbook usage and worth the explicit praise.

**P2 · The "additive, not exclusive" test (line 860-872) explicitly documents an intentional design choice.** The test name ("accepts executable on a non-substrate-construct") + the inline comment ("the conditional only enforces requiredness, not exclusivity") together tell future maintainers what the schema *intends*. Most schema PRs document the rejection paths and leave the permission paths to inference. Calling it out explicitly is good discipline — the next person who adds "exclusivity" via `superRefine` will see this test fail and know to revisit the doctrine.

**P3 · The schema docstrings reference the doctrine (cycle 2026-05-03 substrate-integration) and ground each new shape in OSTROM hexagonal-port discipline + loa-hounfour DomainEvent routing convention.** This is the kind of in-source provenance that pays off six months later when someone tries to remove a "redundant" field — they read the docstring, see the cross-reference, and know to chase down the doctrine before deleting. Far better than the alternative (PR description has the context, code has none).

**P4 · 5 new TypeScript types exported alongside the schemas via `z.infer<typeof X>` (line 687-692).** No hand-written duplication. Consumers who want substrate-shaped types get them from `z.infer` automatically. The drift that exists in F4 above is the *legacy* hand-written interface; the new fields followed the right pattern from the start.

---

### Vercel CI Note

The two `Vercel – constructs-web` and `Vercel – loa-constructs-explorer` failures are pre-existing and unrelated to this PR. The diff is purely backend (`packages/shared/src/validation.ts` + tests) — no frontend code touched, no `apps/explorer/` changes, no `apps/api/` changes. The Vercel deploys would fail on `main` too. Recent merged PRs (#221, #220, #219, #218, #217) all show the same pattern. Safe to ignore for this PR; worth a separate sweep on the deploy config.

---

### Verdict

**Recommendation: MERGE.** The 7 findings above are all "consider before the third instance ships" — none block the construct-lore-essay-grader first-rider. F1 (`runtime: {}` passes) and F2 (`protocol.input/output` optional) are the two worth queueing as immediate follow-up because they're enforcement gaps in the doctrine the PR claims to embody. F4 (drift in `types.ts`) is the queued tech debt that compounds; file an issue. F3, F5, F6, F7 are texture.

The PR does what it says on the tin: extends the manifest with `substrate-construct` archetype + 4 new fields, wires up the conditional, ships 11 tests, preserves backward-compat. Clean unit of work.

— BEAUVOIR (Bridgebuilder)

PR #255, opened 2026-06-05
feat(schemas/runtime): composition v1.4 adds optional Stage.intelligence_tier

zkSoju:
## what

adds an optional `intelligence_tier` (cheap | standard | deep) to a composition Stage, so a `/compose` segment can declare the model-intelligence it needs instead of inheriting the session model. `schema_version` goes to 1.4.

RFC: `grimoires/loa/context/2026-06-03-intelligence-router-coherence.md` (loa-freeside).
runtime consumer: the construct-rooms-substrate PR (segment-emitter resolves the tier to a model).

## change

- `$defs.Stage.intelligence_tier`: optional string enum `[cheap, standard, deep]`, placed beside `thinking_effort` (its sibling cognition knob).
- `schema_version` enum: add `"1.4"` (additive); description note in the existing v1.1/v1.2/v1.3 style.
- `VERSIONING.md`: v1.4 row (and backfilled the missing v1.3 row).
- `additionalProperties: false` on Stage is untouched. this adds a known property, it does not relax validation.

## backward compat

zero new validation errors versus `main` across every existing composition YAML (delta = 0). v1.0 through v1.3 documents validate unchanged.


zkSoju (commented):
## Summary

A disciplined, additive schema extension — the governance hygiene (authorization token, backward-compat promise, optional field) is correctly observed. Two items warrant pre-merge action: a duplicate compatibility claim that silently weakens the stated contract (F-002, confidence 1.0), and an unresolved precedence ambiguity between `intelligence_tier` and `thinking_effort` that will produce consumer divergence (F-003, confidence 0.85).

---

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F-001",
      "title": "VERSIONING.md table skips v1.3 — shows 1.2 → 1.4",
      "severity": "HIGH",
      "category": "documentation",
      "file": ".claude/schemas/VERSIONING.md",
      "description": "The schema registry table was updated from 1.2 to 1.4 in a single diff hunk, but the changelog rows for v1.3 and v1.4 were added in the same PR. The Current column now reads 1.4, but the table previously stopped at 1.2 — meaning v1.3 was never recorded in the table row (only in the changelog). Cross-repo consumers reading the table for the current version would have seen a gap if they consumed between the v1.3 bump and this PR landing.",
      "suggestion": "Confirm v1.3 was published in its own PR with a corresponding table update. If not, add a note that v1.3 and v1.4 are being formally documented here retroactively. The table row should reflect the version at which it was last modified, not just the current tip.",
      "confidence": 0.75,
      "faang_parallel": "Google's protobuf central schema registry enforces a hard rule: every version bump requires a corresponding registry table entry before the CL lands. A gap in the table — even if the changelog row exists — creates a window where consumers reading the registry for the canonical version silently lag. Protobuf tooling reads the table, not the changelog.",
      "metaphor": "The table is the store's shelf label; the changelog is the stockroom receipt. Having the receipt but not the shelf label means consumers don't know the product exists.",
      "teachable_moment": "Schema registries are machine-read as often as human-read. A table gap at the 'current version' column means any automation that reads the table to pin a version will silently under-count. The changelog and the table must advance together — they serve different audiences.",
      "connection": "Connects to the sonar-api-runtime-topology pattern: the authoritative surface must match the deployed state, or consumers diverge from an undetectable starting point."
    },
    {
      "id": "F-002",
      "title": "Description string contains duplicate additive-validation sentence",
      "severity": "LOW",
      "category": "documentation",
      "file": ".claude/schemas/runtime/composition.schema.json",
      "description": "The updated `schema_version` description ends with both '...all additive, v1.0–v1.3 YAMLs validate unchanged.' and then 'All additive; v1.0/v1.1/v1.2 YAMLs validate unchanged.' — the second sentence is a leftover from the v1.3 description and now states a weaker (v1.0–v1.2 only) compatibility claim immediately after a stronger (v1.0–v1.3) one.",
      "suggestion": "Remove the trailing 'All additive; v1.0/v1.1/v1.2 YAMLs validate unchanged.' sentence. The preceding v1.4 sentence already covers the full range.",
      "confidence": 1.0,
      "faang_parallel": "Stripe's API documentation team treats compatibility claims as load-bearing prose with a standing lint rule: a later, weaker claim in the same paragraph is read as an override by careful consumers and their SDK generators. Stripe catches this class of doc contradiction in automated lint before review.",
      "metaphor": "A warranty card that says 'covered for 3 years' in the headline and 'covered for 2 years' in the fine print. In a contract dispute, the fine print wins. Consumers reading carefully will anchor to the weaker claim.",
      "teachable_moment": "In schema descriptions, every sentence is a contract clause. Stale clauses left from copy-paste aren't cosmetic — they're contradictions that erode trust in the whole document. The discipline is: when you bump a version, delete the prior version's sentences, not just append.",
      "connection": "Same failure class as CI sensors that go numb: stale documentation creates false confidence that is strictly worse than no documentation."
    },
    {
      "id": "F-003",
      "title": "No mutual-exclusion or precedence rule between intelligence_tier and thinking_effort",
      "severity": "MEDIUM",
      "category": "schema-design",
      "file": ".claude/schemas/runtime/composition.schema.json",
      "description": "v1.1 introduced `thinking_effort` (a stage-level override) and v1.4 introduces `intelligence_tier`. Both influence model selection. The schema permits both fields to be set simultaneously on a single stage with no defined precedence, creating an ambiguous runtime contract. The description says 'absent → default-by-role' but is silent on the co-presence case.",
      "suggestion": "Add a schema-level `description` note (or a `$comment`) on `intelligence_tier` stating explicitly which field wins when both are present, or use a JSON Schema `oneOf`/`not` to make co-presence a validation error if they are truly mutually exclusive.",
      "confidence": 0.85,
      "faang_parallel": "CSS specificity rules were defined precisely because early browsers had no precedence definition for conflicting declarations — the result was browser-specific rendering behavior that no author intended. JSON Schema has the same footgun: two fields that both mean 'pick a model' without precedence will produce runtime divergence across consumers, each of which makes a different reasonable assumption.",
      "metaphor": "Two co-pilots with hands on the controls and no protocol for who yields. Fine in cruise. A problem at the moment of disagreement.",
      "teachable_moment": "The cost of defining precedence at schema authoring time is one sentence in a $comment. The cost of leaving it undefined is N consumer implementations, each making a different assumption, none of which are wrong because the contract didn't say. Define precedence when you introduce the second field; retrofitting it after consumers ship is a breaking change disguised as a clarification.",
      "connection": "This is the schema-level expression of the schema-is-not-the-contract doctrine: shape is defined here, but behavioral precedence is not — and behavior is the harder half of the contract to retrofit."
    },
    {
      "id": "F-004",
      "title": "Model mapping (cheap→haiku, standard→sonnet, deep→opus) is baked into description prose, not schema",
      "severity": "MEDIUM",
      "category": "schema-design",
      "file": ".claude/schemas/runtime/composition.schema.json",
      "description": "The concrete model mapping lives only in the field description string. When the underlying model lineup changes (e.g., haiku is superseded), the mapping in prose becomes stale silently — nothing in the schema enforces or references a versioned mapping table.",
      "suggestion": "Reference a canonical model-routing table (e.g., a separate `model-tier-map.json` or a section in VERSIONING.md) from the description rather than inlining the mapping. This makes the mapping a single source of truth that can be updated without a schema version bump.",
      "confidence": 0.8,
      "faang_parallel": "AWS IAM managed policy documents reference a versioned policy document ARN, not inline strings. This is exactly why: inline strings in policy descriptions have no owner, no CI gate, and no version. The policy document is auditable and replaceable; the inline string is a copy that diverges. AWS learned this after the early IAM documentation became the de facto spec for third-party tooling and then rotted.",
      "metaphor": "Writing the train schedule on the ticket instead of posting it at the station. Every ticket becomes a lie the moment the schedule changes — and you can't recall the tickets.",
      "teachable_moment": "A description string in JSON Schema has no version, no CI gate, and no diff history beyond 'the string changed.' Any mapping table embedded in prose is a copy that will silently lag the source of truth. The fix is a reference, not a copy — and references compose with tooling in ways that prose does not.",
      "connection": "Mirrors the module-self-distribution-doctrine: each artifact owns its distribution channel. The schema owns shape validation; the routing table owns the mapping. They shouldn't be collapsed into one artifact."
    },
    {
      "id": "F-005",
      "title": "RFC reference in description is not a stable locatable artifact",
      "severity": "LOW",
      "category": "documentation",
      "file": ".claude/schemas/runtime/composition.schema.json",
      "description": "The description cites 'per RFC 2026-06-03-intelligence-router-coherence' as the authorization source for v1.4. RFC documents cited in schema descriptions should be resolvable; if this RFC lives in a grimoire or vault path, the path is not included.",
      "suggestion": "Add the repo-relative path to the RFC document (e.g., `grimoires/loa/context/2026-06-03-intelligence-router-coherence.md`) so auditors can locate the authorization artifact directly.",
      "confidence": 0.7,
      "faang_parallel": "IETF RFC citations work because every RFC resolves to a stable URL via the RFC number. Internal 'RFC' naming without a stable path is naming theater — it signals rigor without providing the property that makes rigor useful. Google's internal design doc culture requires a Buganizer or doc number in every CL; the name alone doesn't pass review.",
      "metaphor": "Citing 'the meeting where we decided this' without a meeting notes link. Future you — or a new team member six months out — cannot find it. The citation exists to make the past findable, not to signal that a past exists.",
      "teachable_moment": "Authorization artifacts cited in schema descriptions are governance breadcrumbs. A breadcrumb that doesn't lead anywhere is decoration that erodes trust in the breadcrumbs that do lead somewhere. Add the path; it costs one edit and saves every future auditor the same search.",
      "connection": "Mirrors the acvp-capability-authorization doctrine: the authorization must be verifiable, not just named. A name without a path is a claim without proof."
    },
    {
      "id": "F-006",
      "title": "Changelog row for v1.3 correctly added; authorization token present",
      "severity": "PRAISE",
      "category": "versioning",
      "file": ".claude/schemas/VERSIONING.md",
      "description": "The v1.3 changelog row includes the cycle number, the operator authorization token (C053.OP-S1), and the behavioral contract ('third seam class'). This is the correct pattern for governed schema changes and makes the audit trail complete.",
      "suggestion": "No action required.",
      "confidence": null,
      "faang_parallel": "Google's CL culture requires every API-affecting change to cite a design doc number. The authorization token pattern here achieves exactly that: the change is traceable to a human decision artifact, not just a diff timestamp. This is what distinguishes a governance trail from a version history.",
      "metaphor": "A chain of custody on evidence — each handoff is signed. The token is the signature. Without it, you have a chain of possession; with it, you have a chain of custody.",
      "teachable_moment": "Authorization tokens in changelogs transform a schema registry from a historical record into an auditable governance artifact. 'Who approved this change and under what authority?' becomes answerable in two seconds instead of two days of git archaeology."
    },
    {
      "id": "F-007",
      "title": "intelligence_tier enum values are stable and non-overlapping with existing thinking_effort",
      "severity": "PRAISE",
      "category": "schema-design",
      "file": ".claude/schemas/runtime/composition.schema.json",
      "description": "The three-value enum (cheap|standard|deep) is distinct from the thinking_effort vocabulary, uses lowercase-hyphenless tokens consistent with other enums in this schema, and is fully optional — existing compositions validate without change.",
      "suggestion": "No action required.",
      "confidence": null,
      "faang_parallel": "Kubernetes resource requests/limits use orthogonal vocabulary deliberately — 'requests' and 'limits' are distinct namespaces even though both influence scheduling. Clean vocabulary separation is what lets consumers reason about each field independently and write tooling that doesn't need to understand both fields to process either.",
      "metaphor": "Two dials on a mixing board, clearly labeled for different instruments. You can turn both without the channels fighting each other — until the routing table upstream conflates them, which is exactly what F-003 flags.",
      "teachable_moment": "Enum vocabulary is a public API surface that outlives the implementation. Tokens that don't collide with existing vocabulary in the same object let consumers grep and type-check each field independently. The 'cheap|standard|deep' choice will age well because it describes intent, not implementation — even if the underlying model names change."
    }
  ]
}
```


---

## Callouts

**F-006 — Authorization token pattern executed correctly.** The v1.3 row (cycle + operator token + behavioral contract in one line) is governance done right. This is what makes "who approved this, when, and under what authority" a two-second lookup rather than a git archaeology expedition. Carry this pattern forward to every governed schema change — it compounds in value the moment a second person needs to audit the trail.

**F-007 — Vocabulary discipline held.** `cheap|standard|deep` is orthogonal to `thinking_effort`, uses tokens that describe intent rather than implementation, and is fully optional with a sensible absent-means-default contract. The vocabulary choice will outlast the current model lineup — which is exactly what you want from an enum that lives in a schema.

---

*The F-003 / F-004 pair is worth reading together: the enum vocabulary is clean, but the routing logic and precedence rules that give the enum meaning are in prose and undefined respectively. The field is well-shaped; the behavioral contract around it needs one more pass.*

PR #264, opened 2026-06-10
fix(schema): persona docstring example cites the eradicated phantom CODEX-RESCUE

zkSoju:
Estate ghost sweep (post construct-compositions#21) is clean except this docstring: the persona-handle schema cites `'CODEX-RESCUE'` as an observed hyphenated value. That persona was a phantom (no identity file anywhere; construct-compositions#20) and is now stripped estate-wide. Swapped for `SCAR-SENTINEL`, a real hyphenated persona in the current census. Related: #263 (fagan unregistered).

### DIFF
--- a/.claude/schemas/runtime/composition.schema.json
+++ b/.claude/schemas/runtime/composition.schema.json
@@ -394,7 +394,7 @@
         "persona": {
           "type": "string",
           "minLength": 1,
-          "description": "Optional persona handle. v1.0/v1.1 convention: uppercase (e.g. 'ALEXANDER', 'MINT'). v1.2 relaxes to free-form string — registry has lowercase ('codex'), hyphenated ('CODEX-RESCUE'), and prose-tagged values."
+          "description": "Optional persona handle. v1.0/v1.1 convention: uppercase (e.g. 'ALEXANDER', 'MINT'). v1.2 relaxes to free-form string — registry has lowercase ('codex'), hyphenated ('SCAR-SENTINEL'), and prose-tagged values."
         },
         "mode": { "$ref": "#/$defs/Mode" },
         "domain": {


zkSoju (commented):
All changes in this PR are Loa framework files. No application code changes to review. Override with `loa_aware: false` to review framework changes. (marker cleared for fable-headless re-review)

zkSoju (commented):
## Summary

A single-token docstring edit that removes a citation of the eradicated CODEX-RESCUE persona from the composition schema. Validation semantics are untouched — no type, constraint, or `$ref` changes — so the risk surface is purely documentary. The one residual concern: the fix replaces one registry-coupled example with another, leaving the drift mechanism that produced this bug fully intact.

The architectural observation worth pausing on: schema docstrings here serve as evidence ("the registry has X, Y, Z") rather than illustration ("values like X are valid"). Evidence-style docs inherit the mutability of what they cite. This PR is the proof — it exists only because a previously true citation became false.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F-001",
      "title": "Replacement example 'SCAR-SENTINEL' should be verified to exist in the registry",
      "severity": "LOW",
      "category": "documentation",
      "file": ".claude/schemas/runtime/composition.schema.json",
      "description": "The docstring asserts a factual claim about current registry contents ('registry has ... hyphenated (SCAR-SENTINEL)'). The diff alone does not demonstrate that SCAR-SENTINEL is a real persona handle in the registry. If the replacement was chosen as merely a plausible hyphenated name rather than a verified registry entry, the docstring would again cite a phantom — the exact defect class this PR fixes.",
      "suggestion": "Confirm 'SCAR-SENTINEL' appears in the live registry (e.g. grep the registry/manifest sources). If it does, no action needed; consider noting the verification in the PR description for the audit trail.",
      "confidence": 0.4,
      "metaphor": "Fixing a phone book entry by writing in a new number you haven't dialed — the correction is only as good as the verification behind it.",
      "teachable_moment": "A fix for a stale-reference bug must itself be verified fresh, or it just resets the staleness clock. Cannot be confirmed from this diff alone; one grep closes the question."
    },
    {
      "id": "F-002",
      "title": "Docstring couples schema to mutable registry contents — same drift will recur",
      "severity": "LOW",
      "category": "maintainability",
      "file": ".claude/schemas/runtime/composition.schema.json",
      "description": "The persona description embeds concrete registry values ('codex', 'SCAR-SENTINEL', prose-tagged values) as evidence for the v1.2 free-form relaxation. Registry contents are mutable, so any cited handle can be eradicated later, reintroducing a phantom reference — this PR is itself proof of that failure mode. The schema's validation behavior (free-form string, minLength 1) does not depend on which specific handles exist.",
      "suggestion": "Rephrase the example to be illustrative rather than registry-factual, e.g. 'v1.2 relaxes to free-form string — lowercase, hyphenated, and prose-tagged values are all valid', or use explicitly synthetic placeholders ('example-persona'). This removes the need to re-edit the schema when registry personas are renamed or removed.",
      "confidence": 0.7,
      "faang_parallel": "Go's testable Example functions are compiled and run by `go test` precisely because prose examples rot; Kubernetes generates API reference docs from source for the same reason. Where no CI binds docs to reality, docs should cite invariants, not instances.",
      "metaphor": "A museum label naming the painting's current owner — every sale makes the label a lie until someone re-engraves it. Label the painting, not the owner.",
      "connection": "F-001 is the acute symptom; this is the chronic condition. Decoupling the docstring from registry state makes F-001-class verification permanently unnecessary."
    },
    {
      "id": "F-003",
      "title": "Surgical, behavior-neutral cleanup of a phantom reference",
      "severity": "PRAISE",
      "category": "documentation",
      "file": ".claude/schemas/runtime/composition.schema.json",
      "description": "The fix is a single-token docstring edit with zero impact on validation semantics — no type, constraint, or $ref changes. Removing references to an eradicated persona keeps the schema's prose consistent with the deprecation work rather than leaving a dangling citation that future readers might treat as authoritative.",
      "suggestion": "None — the minimal scope is correct for this change.",
      "confidence": 0.9,
      "faang_parallel": "Linux kernel patch discipline: one logical change per commit. A reviewer can verify a one-token diff in seconds; bundled cleanups hide both the cleanup and whatever rides along.",
      "teachable_moment": "Completing a deprecation means hunting its citations, not just its definition. Following the eradication into docstrings is the unglamorous half that prevents future readers from resurrecting the phantom."
    }
  ]
}
```


## Callouts

- **[Quality]** The scope discipline is exemplary (F-003): one token, zero semantic drift, traceable to a single intent. This is what makes review trust cheap — because the diff claims nothing beyond what it shows, verification cost is near zero.
- **[Operational]** Chasing a deprecated persona into schema *prose* — not just code paths — shows the right mental model: documentation is part of the deployed surface, and stale docs are live bugs for every agent that reads them as authority.

The meditation this PR earns: it fixes a phantom by citing a new authority, when the durable fix is to stop citing authorities from inside an artifact that no CI re-validates against them. F-002's rephrasing would make this the last PR of its kind for this docstring.

Verdict: **COMMENT**

PR #277, opened 2026-07-15
fix(atlas): verify territory ratification

zkSoju:
## Outcome

Makes `constructs atlas` prove territory ratification with the same fail-closed predicate used by `station`, instead of hardcoding every atlas as `unchecked`. This unblocks the Freeside operator surface `--require-ok` gate while keeping each region root as the authority boundary.

## Changes

- extract shared read-only git/default-branch ratification inspection
- preserve the station mutation gate and error contract
- report per-region verification plus an aggregate atlas status
- keep unreadable or unverified sources fail-closed as `unchecked`

## Verification

- `bun run --cwd packages/constructs-cli test:all` — 177/177 tests, 7/7 deterministic vectors
- merged Freeside consumer + this CLI + installed Beacon/The Arcade packs: `render --json --require-ok` — `status: ok`, territory `ratified`

## E2E defect

The merged consumer correctly rejected the prior producer because `atlas()` hardcoded `ratification_status: unchecked`; the success path was unreachable by construction.

### DIFF
--- a/packages/constructs-cli/lib/ratification.mjs
+++ b/packages/constructs-cli/lib/ratification.mjs
@@ -1,0 +1,140 @@
+// Read-only territory ratification proof shared by station writes and atlas reads.
+// Tool location is never authority: every fact is resolved from the region root.
+
+import path from 'node:path';
+import { git, run } from './exec.mjs';
+import { EXIT } from './contract.mjs';
+
+export const TERRITORY_MANIFEST_REL = path.join('grimoires', 'territory.yaml');
+
+export class RatificationInspectionError extends Error {
+  constructor(message, exitCode = EXIT.CALLER_ERROR, { fix = null } = {}) {
+    super(message);
+    this.name = 'RatificationInspectionError';
+    this.exitCode = exitCode;
+    this.fix = fix;
+  }
+}
+
+export async function defaultBranch(regionRoot) {
+  const origin = await run('git', ['symbolic-ref', '--short', 'refs/remotes/origin/HEAD'], {
+    cwd: regionRoot,
+    allowNonZero: true,
+  });
+  if (origin.exitCode === 0) return origin.stdout.trim().replace(/^origin\//, '');
+  for (const name of ['main', 'master']) {
+    const ref = await run('git', ['show-ref', '--verify', '--quiet', `refs/heads/${name}`], {
+      cwd: regionRoot,
+      allowNonZero: true,
+    });
+    if (ref.exitCode === 0) return name;
+  }
+  return null;
+}
+
+export async function gitFacts(regionRoot, manifestRel = TERRITORY_MANIFEST_REL) {
+  const inRepo = await run('git', ['rev-parse', '--is-inside-work-tree'], {
+    cwd: regionRoot,
+    allowNonZero: true,
+  });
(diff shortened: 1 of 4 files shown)


zkSoju (commented):
## Summary

Verdict: `COMMENT`. This PR’s strongest architectural move is collapsing ratification truth into a shared predicate, then exposing atlas verification at the region level; that turns “trust me” metadata into inspectable evidence. The main review value here is preserving that contract as future ratification rules evolve.

The deeper pattern is provenance discipline: station writes and atlas reads must not develop separate folklore about what “ratified” means. Distributed systems teams learned this repeatedly: once verification rules fork, every downstream consumer starts debugging ghosts instead of state.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "PRAISE-1",
      "title": "Ratification predicate is shared between station and atlas paths",
      "severity": "PRAISE",
      "category": "correctness",
      "file": "packages/constructs-cli/lib/ratification.mjs:86",
      "description": "The ratification checks were extracted into one shared module and reused by both station writes and atlas reads, reducing the chance that the two paths drift on what counts as ratified.",
      "suggestion": "Keep future ratification criteria changes centralized in this module.",
      "confidence": 0.95,
      "faang_parallel": "Google's Borg and later Kubernetes patterns both lean heavily on a single reconciliation predicate: desired state and observed state must be compared through one shared definition, not duplicated local interpretations.",
      "metaphor": "This is like having one official stamp at city hall instead of every department inventing its own seal. The stamp may evolve, but citizens are not asked to guess which seal counts today.",
      "teachable_moment": "Correctness improves when predicates become named domain concepts. Once a rule decides whether state is trustworthy, it deserves a shared API, tests, and vocabulary.",
      "connection": "This connects directly to anti-drift architecture: write paths and read paths often start symmetrical, then quietly diverge. Centralizing the predicate makes future changes auditable."
    },
    {
      "id": "PRAISE-2",
      "title": "Atlas now reports per-region verification details",
      "severity": "PRAISE",
      "category": "observability",
      "file": "packages/constructs-cli/lib/territory.mjs:301",
      "description": "Each region block now exposes ratification status, message, and verification facts, which makes aggregate unchecked states diagnosable instead of only reporting a top-level honesty marker.",
      "suggestion": "Retain this per-region shape as the contract evolves so callers can distinguish local branch, dirty manifest, and remote anchoring failures.",
      "confidence": 0.9,
      "faang_parallel": "Amazon's operational culture favors per-component health over single aggregate lights because aggregate green/red states hide the failing cell. Cell-based systems need local evidence.",
      "metaphor": "A top-level honesty marker is a building fire alarm; useful, but not enough. Per-region verification is the panel that tells firefighters which floor, which sensor, and what changed.",
      "teachable_moment": "Observability should preserve the shape of the system being observed. If territory is regional, verification output should be regional too.",
      "connection": "This strengthens operational readiness: callers can route remediation by failure mode instead of parsing one vague aggregate condition."
    }
  ]
}
```


## Callouts

[Quality] The extraction into `ratification.mjs` is the right kind of abstraction: it removes meaningful duplication while naming a domain rule. That is the standard to keep.

[Operational] The per-region verification details are especially strong because they support future humans and agents. They leave a decision trail: not just “unratified,” but why the system believes that.

[Test Coverage] The added territory tests are a good signal that this behavior is now treated as contract surface, not incidental output formatting.

